PR #6, opened 2016-06-25
support for map operations

jmxpearson:
This will allow mapping a univariate function over an array to be treated/optimized as a single node operation.


src/ReverseDiffPrototype.jl
```diff
@@ -19,12 +19,30 @@ type TapeReal{tag,T<:Real} <: Real
...
+    end
+    out
+end
+@inline function adjoint{tag,T<:Real}(A::Array{TapeReal{tag, T}})
+    out = similar(A, T)
+    for i in eachindex(A)
+        out[i] = A[i].adj
+    end
+    out
+end
```
jrevels: Can we do `A[i].val` --> `value(A[i])` and `A[i].adj` --> `adjoint(A[i])`?

jmxpearson: Oh yeah. That makes sense.

PR #7, opened 2016-06-30
store pointer to parent in TapeNodes instead of pushing to a tape stack

jrevels:
@mlubin So is this what you meant by generating a graph instead of storing a tape? GC time is now a huge factor, since it becomes the GC's job to free the nodes after back-propagation happens.

Using my usual microbenchmark: 

```julia
using ReverseDiffPrototype

function rosenbrock(x)
    a = one(eltype(x))
    b = 100 * a
    result = zero(eltype(x))
    for i in 1:length(x)-1
        result += (a - x[i])^2 + b*(x[i+1] - x[i]^2)^2
    end
    return result
end

x = rand(100000);
out = zeros(x);
g! = ReverseDiffPrototype.gradient(rosenbrock, Input{Float64,100000});
```

On master:

```julia
julia> @time g!(out, x);
  3.108355 seconds (4.50 M allocations: 120.544 MB, 7.68% gc time)
```

With this PR:

```julia
julia> @time g!(out, x);
  0.267648 seconds (3.40 M allocations: 103.759 MB, 53.69% gc time)
```

The target function's runtime:

```julia
julia> using BenchmarkTools

julia> @benchmark rosenbrock($x)
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  0.00 bytes
  allocs estimate:  0
  minimum time:     155.66 μs (0.00% GC)
  median time:      155.80 μs (0.00% GC)
  mean time:        156.89 μs (0.00% GC)
  maximum time:     355.14 μs (0.00% GC)
```

We're still pretty slow vs. the target function ( ~1700x), but a bit faster than we used to be. From preliminary benchmarking, that improvement might get better the higher the input dimension:

```julia
julia> @time g!(out, x) # on master, length(x) = 1,000,000
 31.253529 seconds (45.00 M allocations: 1.177 GB, 3.49% gc time)
```

```julia
julia> @time g!(out, x);
  1.558430 seconds (34.00 M allocations: 1.013 GB, 37.29% gc time)
```

So at `length(x) == 100,000`, this PR demonstrates a ~10x speedup over master,  and at `length(x) == 1,000,000`, this PR demonstrates a ~20x speedup over master.

### DIFF
--- a/src/ReverseDiffPrototype.jl
+++ b/src/ReverseDiffPrototype.jl
@@ -19,7 +19,8 @@ type TapeReal{tag,T<:Real} <: Real
     val::T
     adj::T
 end
-typealias TapeValue{tag, T} Union{TapeReal{tag, T}, Array{TapeReal{tag, T}}}
+
+typealias TapeValue{tag,T} Union{TapeReal{tag,T}, Array{TapeReal{tag,T}}}
 
 TapeReal{tag<:Tag,T}(::Type{tag}, val::T) = TapeReal{tag,T}(val)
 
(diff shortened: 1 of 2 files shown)


jmxpearson:
Cool!

The other advantage of the graph is that one can query it, perform partial evaluations, etc. My understanding is that this is the Theano approach. 

But I was always worried about memory. 

Probably with Array nodes, this becomes much less of a concern.

mlubin:
Yep. Note you can compare with JuMP's AD:
```jl
using JuMP
m = Model()
@variable(m, x[1:100000])
@NLobjective(m, Min, sum{ (1 - x[i])^2 + 100*(x[i+1]-x[i]^2)^2, i=1:length(x)-1})
d = JuMP.NLPEvaluator(m)
MathProgBase.initialize(d, [:Grad])
∇f = zeros(100000)
values = rand(100000)
@time MathProgBase.eval_grad_f(d, ∇f, values)
@time MathProgBase.eval_grad_f(d, ∇f, values)
```

On the second call I get:
```
  0.012159 seconds (9 allocations: 240 bytes)

```

jrevels:
Wow, that's fast. 

Is that time including the full construction/destruction of the graph (i.e. forward pass + back-propagation + GC'ing the graph afterwards)? It doesn't seem like it from the memory measurements. 

Are you caching the graph for subsequent calls, and just reinterpreting it with new values? I don't think we can do that here, since our forward trace needs to accommodate value-dependent branches. 

I was going to ask about how to make our "interpreter" faster, it seems to partly be a game of what to share with the compiler and what to hide from it.

mlubin:
The graph is stored between calls and we just do another pass through it
when the values change, so there's no GC involved for evaluating
derivatives.

PR #10, opened 2016-07-11
Calculate intermediary derivatives during forward pass and optimize/simplify graph layout

jrevels:
This PR changes our approach with the goal of making the backwards pass (nearly) non-allocating. We no longer need external graph storage or caching; the number type itself is now its own node in the graph, and maintains its own references to its parents. Since there's no longer a tape, I've named the new value type `TraceReal`. By using `RefValue`s for the `adjoint` field, I was able to make `TraceReal`s immutable.

The backwards pass for gradients now does a BFS starting at the result node. I use the normal queue-based implementation of a BFS, which means there is a small amount of allocation, but it's pretty negligible compared to allocation of the forward pass:

```julia
julia> include("test/rosenbrock.jl")
  0.058133 seconds (3.20 M allocations: 96.138 MB) # forward pass only
  0.154665 seconds (10 allocations: 896 bytes) # backward pass only
  0.214749 seconds (3.20 M allocations: 96.130 MB) # call to gradient! (forward + backward pass)
done
```

The majority of the time (~ 70%) of the backwards pass is graph traversal, rather than derivative propagation. Here's the same run, with the derivative propagation commented out:

```julia
julia> include("test/rosenbrock.jl")
  0.058232 seconds (3.20 M allocations: 96.130 MB)  # forward pass only
  0.128279 seconds (10 allocations: 896 bytes) # backward pass only, no derivative propagation
  0.198831 seconds (3.20 M allocations: 96.130 MB) # call to gradient! (forward + backward pass)
done
```

 To speed up the BFS, I played around with `DataStructures.Queue`, as well as rolling my own fixed-size queue using a custom ring buffer type, but neither were as performant as just `pop!`/`unshift!` on `Vector`. The queue never exceeds a size of ~20 for rosenbrock, so the `Vector` probably reaches full size via dynamic allocation rather quickly in most cases.

Inspired by @mlubin's approach in ReverseDiffSparse, intermediary derivatives are now calculated in the forward pass using the `ForwardDiff.Dual` type, which means the backwards pass is merely multiplication/addition. I haven't yet explored how far I could go with ReverseDiffSparse's traversal optimizations, which rely on the target function being data independent (i.e. no value-based branching) and representable as a tree (correct me if I'm wrong, Miles). It would be cool to optionally allow the user to annotate things or provide keyword arguments that would make these optimizations possible; I'll open an issue for this.

@jmxpearson I've also added you to the license.

I still have to figure out the right way to do array operation optimizations (e.g. `map` overloading) using this new graph structure.


jrevels:
> I still have to figure out the right way to do array operation optimizations (e.g. map overloading) using this new graph structure.

My naive first thought on this is that we'll have to implement a `TraceArray` wrapper for `AbstractArray`s.

jmxpearson:
Thanks. I finally got the big PR we'd been working on settled, so I'll have
more time to take a look at this.

jmxpearson:
This looks great!

A couple of questions:

1. Does this approach rule out the matrix optimizations we have on master? 

    From @mlubin, I understand that in the case of JuMP, there are often huge numbers of variables with thousands of constraints, all distinct. But for many of the ML applications I want, the graph is not terribly large; many of the nodes, though are substantial tensors. The approach of using an elementwise tape on these will be pretty inefficient, I think, since it prevents Julia from doing things like using optimized linear algebra for backprop.

2. I've thought once again about trying to spec out the codegen approach, this time allowing variables to be arrays in addition to scalars. This should cut down generated code a lot and avoid the bloat I feared initially. Have you thought much about this? Is there something simple I'm missing?

jrevels:
> Does this approach rule out the matrix optimizations we have on master?

I've been thinking hard about this one as well. If we were to continue with this approach, we'd need different wrapper types for the different functions we'd like to optimize. Specifically, I'm thinking about `TraceArray <: AbstractArray` for array-to-array functions (like `map`) and `TraceReduction <: AbstractTraceReal <: Real` for array-to-scalar operations (like `sum`). Then, we'd implement `backprop_step!` on these types. 

I do wonder if there's a way to combine this PR's approach and the approach on `master`, i.e. leveraging `immutable` + `Ref` + forward derivative accumulation with `Dual`s to achieve a non-global tape-based storage that is easier to traverse and optimize array methods for. This thought is what's kept me from hitting merge here.

> I've thought once again about trying to spec out the codegen approach, this time allowing variables to be arrays in addition to scalars. This should cut down generated code a lot and avoid the bloat I feared initially. Have you thought much about this? Is there something simple I'm missing?

I've only thought about the compiled approach a little bit versus the interpreted approach, so I have no idea what the hurdles would be there. Sounds like it's worth trying, though.

I was thinking another big win for the compiled approach could come from partitioning the DAG into smaller subsections that could be compiled as smaller Julia functions, which would probably lead to better compile-time and run-time performance. We could accumulate info about function barriers during the trace that might make this feasible. The real trick with the compiled approach would be not re-compiling for data-independent (i.e. tree-representable) subsections of the DAG. With that in mind, it might be that the compiled approach could only work well for the case described in #11.

jrevels:
superceded by #13

PR #12, opened 2016-07-19
add optimizations of matrix ops

jmxpearson:
I decided to do some timing of my original objective function involving some matrix ops for comparison. By adding some specific rules for handling `*`, I got a 50x speedup over master, which is not entirely surprising. It does suggest, though, that we might want to be sure that approaches like #10 allow for these same kinds of optimizations. My intuition, again, is that unlike in JuMP, a lot of the applications ML people have in mind consist of pretty small graphs with "fat" nodes.

Timing:
(master, without matrix ops)
```
0.179432 seconds (6.17 M allocations: 188.689 MB)
20100-element Array{Float64,1}
```

this branch:
```
0.003167 seconds (80.46 k allocations: 3.448 MB)
20100-element Array{Float64,1}
```

### DIFF
--- a/src/ReverseDiffPrototype.jl
+++ b/src/ReverseDiffPrototype.jl
@@ -40,6 +40,21 @@ typealias TapeArray{F,T,N} AbstractArray{TapeReal{F,T},N}
 @inline numtype{F,T}(::Type{TapeReal{F,T}}) = T
 @inline value(n::TapeReal) = n.val
 @inline adjoint(n::TapeReal) = n.adj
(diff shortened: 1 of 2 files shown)


jrevels:
> we might want to be sure that approaches like #10 allow for these same kinds of optimizations.

Definitely. This LGTM, merge when ready.

jmxpearson:
I'm going to try something similar for your simplify branch, since that's even faster to start. It may take me a couple of days to get to it, though.

Issue #3, opened 2016-06-25
Replace Array{TapeReal} with TapeArray


jmxpearson:
We want this so that custom Array types that don't subtype AbstractArray can work. Also, arrays of distributions with TapeReal parameters. 

Possible API: `wrap` and `unwrap` to convert Arrays of TapeReals to TapeArrays and vice-versa.

jrevels:
Closing since `TrackedArrays` now exist.

PR #13, opened 2016-08-09
Optimize tracing implementation, utilize hybrid approach for intermediary derivative calculation

jrevels:
This PR supercedes #10.

Basically, this PR cleans up a bunch of code, as well as combines the current `master` approach with the advantages of #10. Function traces are now stored in a custom linked-list style data structure, which speeds up backpropagation significantly, even more so than in #10. Thus, the forward-pass is now the bottleneck.

By default, like in #10, intermediary derivatives are calculated during the forward pass. This works really well for low-dimensional functions. However, keeping the trace-based approach allows us to overload specific operations so that their derivative calculations are postponed until the reverse pass, allowing us to easily implement the kind of matrix optimizations found in #12. As proof that this works - and so as not to introduce regressions - I've reimplemented #12 for `map`, `*`, `+`, and `-`. EDIT: I just threw in optimizations for multi-argument `map` and `broadcast`, as well.

Once this PR is merged, I think we'll have a solid foundation over which we can implement a more substantial API (i.e. implement more methods besides`gradient!`). This implementation is also amenable to the optimization proposed in #3, which would significantly reduce the number of unnecessary temporary arrays we currently use for matrix optimizations. 

Obligatory benchmark results:

```julia
julia> include("test/benchmark.jl")
benchmarking rosenbrock(x)...
  0.187755 seconds (3.70 M allocations: 108.337 MB, 80.99% gc time)
  0.052970 seconds (6 allocations: 208 bytes)
  0.234658 seconds (3.70 M allocations: 108.337 MB, 61.86% gc time)
done
benchmarking ackley(x)...
  0.078708 seconds (2.20 M allocations: 62.562 MB, 68.21% gc time)
  0.014728 seconds (6 allocations: 208 bytes)
  0.097820 seconds (2.20 M allocations: 62.563 MB, 54.64% gc time)
benchmarking matrix_test(x)...
  0.000821 seconds (36.37 k allocations: 1.302 MB)
  0.001835 seconds (30 allocations: 226.219 KB)
  0.002365 seconds (36.40 k allocations: 1.523 MB)
done
```


jmxpearson:
Very cool. Excited to take a look!

jrevels:
@jmxpearson Mind if I merge this?

jmxpearson:
Please do!

PR #17, opened 2016-08-16
add jacobian methods and fix #16

### LINKED ISSUE #16: Our higher-order function optimizations don't work if the passed-in function closes over TraceReals
```julia
julia> using ReverseDiffPrototype; const RDP = ReverseDiffPrototype;

julia> function f(x)
           a = one(eltype(x))
           return sum(map(i -> a * i, x))
       end
f (generic function with 1 method)

julia> RDP.gradient!(zeros(3), f, rand(3)) # wrong, should be all ones
3-element Array{Float64,1}:
 0.823614
 0.459833
 0.581877
```

Note that we could replace `a = one(eltype(x))` with something like `a = x[1]`, and the bug would persist. 

The closed-over `TraceReal` is causing each application of the closure to be written to the trace, laying down incorrect nodes that get accumulated in the reverse pass.  It's a perturbation-confusion-esque bug, but one that we're inducing on ourselves instead of it being induced by competing differentiation operators. 

Not only does it give the wrong answer, but it also kind of defeats the point of the optimization (to elide the tracing of the closure application).

I'm unsure how we can resolve this short of scrubbing the closure for `TraceReal`s, just as we'd have to do to fix JuliaDiff/ForwardDiff.jl#83. 

@mlubin any ideas?

jrevels:
I had already prepared the jacobian stuff when I found the bug from #16, so I just rolled the proposed macro into this PR.

~~I settled on `@diffpure` as the macro name, since it's a differentiation optimization for pure functions.~~ Changed it to `@fastdiff`, inspired by the existing `@fastmath` in Base. It also means we can use it for other optimizations in the future, since we're not tying it to purity-based optimizations. Feel free to bikeshed the name.

### DIFF
--- a/benchmark/benchmark.jl
+++ b/benchmark/benchmark.jl
@@ -4,15 +4,7 @@ const N = 100000
 
 println("benchmarking rosenbrock(x)...")
 
-function rosenbrock(x)
-    a = one(eltype(x))
-    b = 100 * a
-    result = zero(eltype(x))
-    for i in 1:length(x)-1
-        result += (a - x[i])^2 + b*(x[i+1] - x[i]^2)^2
-    end
-    return result
-end
+rosenbrock(x) = sum(@diffpure map((i, j) -> (1 - j)^2 + 100*(i - j^2)^2, x[2:end], x[1:end-1]))
 
 x = rand(N);
 out = zeros(x);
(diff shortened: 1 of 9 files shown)


jrevels:
Tests I added here require JuliaDiff/ForwardDiff.jl#138 to pass

jrevels:
Tests are passing here, so I'll merge (I had to cancel the OSX build because of [this](https://www.traviscistatus.com/incidents/4mvp857qx8bw), but I'm on OSX and things pass locally).

PR #20, opened 2016-08-25
Better optimization macros

jrevels:
This basically implements my [idea here](https://github.com/jrevels/ReverseDiffPrototype.jl/issues/11#issuecomment-242461350). 

Instead of `@fastdiff`, we now have `@forward` and `@skip`. I've rewritten all of our existing operations to essentially use these mechanisms instead, and it results in (IMO) cleaner code with no performance penalty. More importantly, users can now take advantage of these mechanisms.

#### `@forward`

The user can wrap any function in `@forward` to calculate that function's derivatives in the forward pass using ForwardDiff, without writing intermediary nodes that otherwise would've needed to be recorded. For example, this would normally take three nodes to record:

```julia
(x->sin(x) + cos(x))(t::TraceReal)
```

If we wrap the function with `@forward`,  then it will only take one node to record:

```julia
@forward(x->sin(x) + cos(x))(t::TraceReal)
```

Instead of relying on `@fastdiff`, our `map` and `broadcast` optimizations are now achieved by overloading on the type produced by `@forward(f)`, resulting in a more precise and composable syntax.

#### `@skip`

Wrapping a function in `@skip` causes calls to that function to not get recorded to the trace. For an example, I use this mechanism internally to implement comparator methods on `TraceReal`s.


jrevels:
Tests pass, the error is just due to recent OSX outage on Travis.

jmxpearson:
This is all really fantastic work!

jrevels:
Thanks! In addition to the callsite annotation this PR started with, I've added the ability to use the annotations at definition-time, e.g.

```julia
@forward sigmoid(n) = 1. / (1. + exp(-n))
```

OSX builds on Travis are still out of commission, but I run OSX locally and the tests pass. I'll sit on this for a couple of hours and then merge.

jrevels:
Yay, OSX builds are finally back! Looks all green here.

PR #21, opened 2016-08-26
support the ForwardDiffResult types and hessians

jrevels:
I set down to implement Hessian support today, and found out that #19 actually allows for nested differentiation without any extra work. Woohoo!

This PR still needs tests.


mlubin:
If you do some work to detect sparsity patterns, then you can easily compute sparse hessians using the code in ReverseDiffSparse.

jrevels:
Yeah, taking the Jacobian of the gradient is simply the most naive thing possible, so I'd figure it'd be a good starting point. 

I'd love to improve Hessians here by leveraging ReverseDiffSparse's algorithms - how would I go about using it? Could I reuse the graph coloring stuff ReverseDiffSparse implements?

mlubin:
> Could I reuse the graph coloring stuff ReverseDiffSparse implements?

Yes, in principle, but may need some refactoring first. Dunno if it's worth trying to merge with https://github.com/mauro3/MatrixColorings.jl.

jrevels:
I could use graph coloring stuff for sparse Jacobians, as well, yeah? Taking advantage of sparsity patterns is new territory for me.

mlubin:
@jrevels, yep, definitely.

jrevels:
Just an update: I'm going to wait until #22 lands on master, then rebase this branch on top of master. There's still a bug here I've got to fix (maybe more than one 😛).

jrevels:
I ran into some really tricky bugs with this one, mainly involving mixed `Dual` and `TraceReal` operations occurring during the backward pass for certain functions. This PR is stale at this point, so I'm gonna shelve nested differentiation for now. I may figure out a working implementation at a later point. If so, I'll just open a new PR then.

PR #38, opened 2016-12-22
add MNIST example

jrevels:
Here's an MNIST example using ReverseDiff (along with some minor improvements to the package I made along the way). 

My goal was mainly to have a self-contained example to show off the model + gradient definition using ReverseDiff; the real-world batching/loading/training code is something that I would expect to be provided by front-end machine learning libraries (i.e. ReverseDiff dependencies). 

I don't have any actual machine learning background, so this will probably contain some noob mistakes -any onlookers with actual ML experience are welcome to chime in! I also still need to write a test to check if this thing works correctly, unless @ccxxxx or @jmxpearson get to it first (feel free to open a PR against this PR to add the test).


examples/mnist.jl
```diff
@@ -0,0 +1,128 @@
...
+# forward-mode. This allows us to call `minus_log.(y)` instead of `-(log.(y))`. By defining
+# our own "fused" `minus_log` kernel using `@forward`, the operation `minus_log.(y)` becomes
+# a single array instruction in the tape (instead of two separate ones,  as is the case with
+# `-(log.(y))`), buying us a slight performance gain. Note that we didn't *need* to do this;
+# it's simply a good place to show off the `@forward` feature.
+ReverseDiff.@forward minus_log(x::Real) = -log(x)
+
+cross_entropy(y′, y) = mean(sum(y′ .* (minus_log.(y)), 1))
+
+softmax(x) = (exp_x = exp.(x); exp_x ./ sum(exp_x))
```
jrevels: Ah, thanks. 

I'm slightly confused - shouldn't we be normalizing the columns, not the rows, and thus use `sum(exp_x, 1)` here instead of `sum(exp_x, 2)`?  I ask because, the way I've set this up, each column of `pixels` (in `model`) corresponds to an input image, and such that `size(pixels, 2) == BATCH_SIZE` (transposed from the way Tensorflow does it, since Julia is column-major instead of row-major). Intuitively, I would think we want to normalize along each input column instead of along each row in the batch...but like I said, I'm still new to this stuff.

domluna: If the second dimension represents the `BATCH_SIZE` then it would be `sum(exp_x, 1)`. I'm just used to the first dimension being the `BATCH_SIZE` so I kind of just assumed that, should have taken a closer look. For each example/image we'd want to get a probability distribution of what class/label it thinks that image belongs to.

jrevels: Makes sense. I'll add a comment about the transposition, most people are probably used to the row-major way. Thanks again!

mohamed82008:
Closing due to lack of activity and the PR being outdated.

PR #46, opened 2017-01-18
Xin/dev

ccxxxx:
Complete with MNIST example. Run examples/mnistTest.jl to see the good/bad results.

### DIFF
--- a/examples/mnist.jl
+++ b/examples/mnist.jl
@@ -246,7 +246,7 @@ train_all!(∇batch, batch, TRAIN_IMAGES, TRAIN_LABELS)
 # The functin is used to calculate column-wise matches
 function accuracy(result::Array, label::Array)
     score = 0
-    _, ncol = size(label)
+    ncol = size(label, 2)
     for i in 1:ncol
         if indmax(result[:, i]) == indmax(label[:, i])
             score += 1
--- a/examples/mnist.jl
+++ b/examples/mnist.jl
@@ -24,7 +24,7 @@ const TRAIN_IMAGES, TRAIN_LABELS = load_mnist(MNIST.traindata())
 const TEST_IMAGES, TEST_LABELS = load_mnist(MNIST.testdata())
 
 # loading batches #
-#-----------------#
+# -----------------#
 
 immutable Batch{W,B,P,L}
     weights::W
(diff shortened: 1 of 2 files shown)


domluna:
@ccxxxx  I have a sample that works on MNIST https://github.com/domluna/Butterfly.jl/blob/master/src/nn.jl (using the jr/mnist branch). It's a bit rough around the edges but the hyperparams might be helpful.

ccxxxx:
@domluna Hey Dominique, thanks for the comment! Yes the example is very helpful but it creates the same problem - which in my commit I described it as "bad example". The weights of training set will spit out NaNs when the learning rate is too large, although we can still get some kind of "accuracy" on the test images, the result is not reliable. 

The intuition - when the learning rate is too large, it is very likely that the gradient step out the domain of loss function (cross entropy). Theoretically such thing could happen, but Tensorflow, on the other hand, seems to bypass this problem even with a constant huge learning rate and the accuracy is still ~80%. I am still looking into their way of getting around this issue, and I don't think this is the problem of ReverseDiff. As of now we just need to provide a good default value and warn user about the NaNs when the learning rate is chosen inappropriately.

domluna:
> The weights of training set will spit out NaNs when the learning rate is too large

What's the learning rate? I've run the program for rates between 1e-3 and 1 and the accuracy is usually 90%+. I'm curious to see the TF program you're referring to. I think weight initialization and preprocessing have a bigger impact than the learning rate.

ccxxxx:
Ah, thanks for the insight! I preprocessed the data by dividing the pixels by 255 (like you did) and it seems to fix the problem with rate <=10. I can think of several reasons but is there a particular reason or strategy for preprocessing the image data in this nature? 

FYI, the TF example is here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/tutorials/mnist/mnist_softmax.py
I played with the learning rate in both ReverseDiff and Tensorflow. Learning rate determines how fast the algorithm learns to find the global/local minimum. If the learning rate is too large, the algorithm is likely to jump around and cannot settle down. Tensorflow seems to be very robust in terms of learning rate (like a crazy large number of 50 or even 1000). 

And I don't think the weight initialization is important since it get updated constantly, as long as it does not start from some weird places. What do you think? Love to hear more!

mohamed82008:
Closing due to lack of activity and the PR being outdated.

PR #47, opened 2017-01-23
cherry-picked improvements + fixes from jr/mnist (including compat fixes for v0.6)

jrevels:
It makes sense to pull out some of these changes from the jr/mnist branch and merge them separately (after which I'll rebase jr/mnist on top of master). This way, I can focus on some of the compat fixes we need for v0.6 without getting tied up in the MNIST example. 

There's still some convert issue on v0.6 I need to fix, and I also need to tag new versions of Calculus and ForwardDiff (so I can bump the version requirements here).


jrevels:
Getting tests to pass here on v0.6 will require https://github.com/JuliaLang/julia/issues/20200 to get fixed.

jrevels:
This branch contains a grab-bag of useful improvements/fixes. I might merge this soon (as long as tests pass on v0.5) even though v0.6 tests aren't passing because there's not much I can do about JuliaLang/julia#20200 for the time being.

PR #60, opened 2017-03-15
correct the seeding order for seeded_reverse_pass! functions

### LINKED ISSUE #59: ReverseDiff doesn't work for noop f(x)=x
```julia
julia> x_seed = [1.0]
1-element Array{Float64,1}:
 1.0

julia> f(x) = x
f (generic function with 1 method)

julia> gcfg = ReverseDiff.GradientConfig(x_seed)
ReverseDiff.GradientConfig

julia> g! = (x, out) -> ReverseDiff.gradient!(out, f, x, gcfg)
(::#17) (generic function with 1 method)

julia> g!([1.0], [0.0])
1-element Array{Float64,1}:
 0.0

julia> f(x) = 1.0*x
f (generic function with 1 method)

julia> g!([1.0], [0.0])
1-element Array{Float64,1}:
 1.0

julia> f(x) = 3.0*x
f (generic function with 1 method)

julia> g!([1.0], [0.0])
1-element Array{Float64,1}:
 3.0
```
Granted, not the most interesting function in the world, but there must be something that you're missing here, since it doesn't think the function changes with `x`.

jrevels:
This fixes #59. Test is added [here](https://github.com/JuliaDiff/DiffBase.jl/pull/3). We'll have to re-run Travis once that's merged and tagged to get representative testing in this PR.

Funnily, enough, this also reveals similar bugs in both ForwardDiff and Calculus - both of them appear to get the wrong answer for the Jacobian of the identity. I've checked locally that this PR enables ReverseDiff to get the correct answer, but once the DiffBase PR is merged, tests will fail here, since ReverseDiff is checking against ForwardDiff, which checks against Calculus.

I'll start working on bug fixes for those packages as well.

### DIFF
--- a/src/api/utils.jl
+++ b/src/api/utils.jl
@@ -14,10 +14,10 @@ function seeded_reverse_pass!(result::AbstractArray, output::AbstractArray, inpu
 end
 
 function seeded_reverse_pass!(output::TrackedReal, input::TrackedReal, tape)
-    seed!(output)
+    pull_value!(output)
     unseed!(input)
+    seed!(output)
     reverse_pass!(tape)
-    unseed!(output)
     return deriv(input)
 end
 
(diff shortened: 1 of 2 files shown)


jrevels:
Tests should pass here once https://github.com/JuliaLang/METADATA.jl/pull/8357 is merged

pkofod:
Rerun tests? PR is merged at metadata

PR #66, opened 2017-04-03
add similar for tracked array

SimonDanisch:


jrevels:
Thanks! Can you add some tests [somewhere around here](https://github.com/JuliaDiff/ReverseDiff.jl/blob/master/test/TrackedTests.jl#L623) checking `similar` against a couple of element type/array type/dimensions combinations?

PR #71, opened 2017-05-24
WIP: replace eval()-ed methods with FunctionWrappers.jl

### LINKED ISSUE #70: World age issues with compiled tapes on 0.6
Compiled tapes work fine from the REPL, but trying to construct one inside a function results in everyone's favorite 0.6 error: 

```julia
julia> using ReverseDiff

julia> function foo(f, x)
           tape = ReverseDiff.compile(ReverseDiff.GradientTape(f, x))
           result = similar(x)
           ReverseDiff.gradient!(result, tape, x)
       end
foo (generic function with 1 method)

julia> g = y -> sum(x -> x^3 + x^2 - x - sin(x), y)
(::#3) (generic function with 1 method)

julia> foo(g, rand(10))
ERROR: MethodError: no method matching forward_pass!(::ReverseDiff.CompiledTape{Symbol("##657"),ReverseDiff.GradientTape{##3#5,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void}}})
The applicable method may be too new: running in world age 21811, while current world is 21814.
Closest candidates are:
  forward_pass!(::ReverseDiff.CompiledTape{Symbol("##657"),ReverseDiff.GradientTape{##3#5,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void}}}) at /Users/rdeits/.julia/v0.6/ReverseDiff/src/api/tape.jl:84 (method too new to be called from this world context.)
  forward_pass!(::ReverseDiff.CompiledTape{Symbol("##656"),ReverseDiff.GradientTape{#f,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void}}}) at /Users/rdeits/.julia/v0.6/ReverseDiff/src/api/tape.jl:84
  forward_pass!(::Array{ReverseDiff.AbstractInstruction,1}) at /Users/rdeits/.julia/v0.6/ReverseDiff/src/tape.jl:76
  ...
Stacktrace:
 [1] seeded_forward_pass! at /Users/rdeits/.julia/v0.6/ReverseDiff/src/api/tape.jl:42 [inlined]
 [2] gradient!(::Array{Float64,1}, ::ReverseDiff.CompiledTape{Symbol("##657"),ReverseDiff.GradientTape{##3#5,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void}}}, ::Array{Float64,1}) at /Users/rdeits/.julia/v0.6/ReverseDiff/src/api/gradients.jl:79
 [3] foo(::Function, ::Array{Float64,1}) at ./REPL[2]:4

julia> versioninfo()
Julia Version 0.6.0-rc2.0
Commit 68e911be53 (2017-05-18 02:31 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin13.4.0)
  CPU: Intel(R) Core(TM) i7-2860QM CPU @ 2.50GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, sandybridge)
```

(this is with ReverseDiff master)

rdeits:
Since this seems like it might not be a terrible idea, I'm opening a PR to figure out whether this is a good change.

As discussed in #70, using `eval()` to generate the forward and reverse pass methods inside `compile()` creates world age issues, which ultimately means that a tape cannot be compiled and used in the same runtime context. This PR fixes that problem by removing the `eval()`ed methods entirely. Instead, to avoid type instability, the compiled tape is augmented with two vectors of concretely typed FunctionWrappers. Each wrapper is built around a functor which, when called, invokes its corresponding instruction's `forward_exec!` or `reverse_exec!` method. 

Performance seems to be ~15% slower on my machine using this branch vs `master` for the rosenbrock example from #70. Hopefully we can improve that...

I've kept the `compile()` method intact for now. `compile()` is *much* faster than before (135ms reduced to ~~8ms~~ 2ms for the rosenbrock example) but it's still not free. 

~~This requires FunctionWrappers master. I've opened an issue there to ask for a new tag: https://github.com/yuyichao/FunctionWrappers.jl/issues/3~~

fixes #70

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ notifications:
 script:
  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
  - julia -e 'Pkg.init(); Pkg.add("FunctionWrappers"); Pkg.checkout("FunctionWrappers")'
- - julia -e 'Pkg.clone(pwd()); Pkg.build("ReverseDiff"); Pkg.test("ReverseDiff"; coverage=true)'
+ - julia -e 'Pkg.clone(pwd()); Pkg.build("ReverseDiff"); Pkg.test("ReverseDiff"; coverage=VERSION >= v"0.6-")'
 
 after_success:
   # push coverage results to Coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,9 @@ notifications:
   email: false
 script:
  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+ - julia -e 'Pkg.init(); Pkg.add("FunctionWrappers"); Pkg.checkout("FunctionWrappers")'
  - julia -e 'Pkg.clone(pwd()); Pkg.build("ReverseDiff"); Pkg.test("ReverseDiff"; coverage=true)'
+
 after_success:
   # push coverage results to Coveralls
   - julia -e 'cd(Pkg.dir("ReverseDiff")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
   - linux
 julia:
   - 0.5
+  - 0.6
   - nightly
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ notifications:
   email: false
 script:
  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
- - julia -e 'Pkg.init(); Pkg.add("FunctionWrappers"); Pkg.checkout("FunctionWrappers")'
  - julia -e 'Pkg.clone(pwd()); Pkg.build("ReverseDiff"); Pkg.test("ReverseDiff"; coverage=VERSION >= v"0.6-")'
 
 after_success:
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,4 @@ julia 0.5
 DiffBase 0.0.3 0.3.0
 ForwardDiff 0.3.4 0.5.0
 Compat 0.19.0
+FunctionWrappers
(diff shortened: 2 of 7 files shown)


jrevels:
@yuyichao, do you feel that FunctionWrappers is robust enough for use by downstream packages like ReverseDiff? I'm asking because you originally referred to it as a [proof-of-concept implementation](https://github.com/JuliaLang/julia/issues/13984).

cc @vtjnash, since we once had a discussion about using [your proposed Callback type](https://discourse.julialang.org/t/proposal-for-a-first-class-dispatch-wrapper/1127) to basically do what this PR does. Ideally, we'll eventually have Base support for dispatch wrappers, but there does seem to be some debate regarding what they'll look like.

vtjnash:
I think that FunctionWrappers is somewhat tedious to use, which is why I want to go with my Callback proposal, but I have nothing against its functionality.

rdeits:
Ok, I've cleaned things up a bit and added some comments. One of the builds with Julia v0.5 stalled on Travis during the nested jacobians test, but I think that's just a slow test (that test set took almost 20 minutes during the last build on master). Turning off code coverage for Julia v0.5 fixed it (and cut the test time in half).

jrevels:
> One of the builds with Julia v0.5 stalled on Travis during the nested jacobians test, but I think that's just a slow test (that test set took almost 20 minutes during the last build on master).

Yeah, that test is really rough on the compiler. 

I ran `benchmark/benchmark.jl` on both master and this PR, and there don't seem to be any notable slowdowns, so I'll merge this tomorrow unless there are any objections.

jrevels:
Thanks @rdeits!

PR #75, opened 2017-06-19
do not rely on map promotion for capturing

jrevels:
fixes the bug exposed by https://github.com/JuliaLang/julia/pull/22398 (thanks @tkelman)


tkelman:
nightly failure looks unrelated

PR #83, opened 2017-07-17
fix overly-inclusive convert method at supertype boundary

### LINKED ISSUE #82: ReverseDiff.gradient incorrect with `if` and independent values.
I *think* this is a mistake (and maybe related to trying to do ReLu in #77?

When I tried to find the gradient for a function that sometimes doesn't use one of its inputs, instead of just that one input's gradient turning to 0, the gradients for all the elements of the input vector turned to 0:
```julia
julia> f(X) = sum([if x > 0; x else 0 end for x in X])
>> f (generic function with 1 method)

julia> ReverseDiff.gradient(f, [2,3,-1])
>> 3-element Array{Int64,1}:
 0
 0
 0

julia> ReverseDiff.gradient(f, [2,3,1])
>> 3-element Array{Int64,1}:
 1
 1
 1
```
I think this is a mistake? Only the third element in the returned array should be 0, since the other two variables *do* have a linear effect on the output of `f`.

Also, though, it currently does get this right for `max(0,x)`, which is another way to implement the same thing (I guess it considers `max(0,x)` inclusive for `x==0`):
```
julia> f(X) = sum([max(0, x) for x in X])
>> f (generic function with 1 method)

julia> ReverseDiff.gradient(f, [2,3,0])
>> 3-element Array{Int64,1}:
 1
 1
 1

julia> ReverseDiff.gradient(f, [2,3,-1])
>> 3-element Array{Int64,1}:
 1
 1
 0

```

jrevels:
Fixes #82. The former "correct" behavior was actually dependent on a type inference bug, which got fixed in the final Julia v0.6.0 release.

Test added in https://github.com/JuliaDiff/DiffBase.jl/pull/12, so I'm going to wait to merge this until that gets tagged (so I can re-run Travis here with the test).

cc @NHDaly

EDIT: Also made nightly failures allowable, since the plan is to replace this package by the v0.7 release anyway

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.5
-DiffBase 0.0.3 0.3.0
+DiffBase 0.0.3 0.4.0
 ForwardDiff 0.3.4 0.5.0
 Compat 0.19.0
 FunctionWrappers 0.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ julia:
   - 0.5
   - 0.6
   - nightly
+matrix:
+  allow_failures:
+    - julia: nightly
 notifications:
   email: false
 script:
--- a/src/api/hessians.jl
+++ b/src/api/hessians.jl
@@ -47,7 +47,7 @@ function hessian!(result::DiffResult, f, input::AbstractArray,
     ∇f! = (y, x) -> begin
         gradient_result = DiffResult(zero(eltype(y)), y)
         gradient!(gradient_result, f, x, cfg.gradient_config)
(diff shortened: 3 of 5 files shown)


NHDaly:
+1 thanks! :)

REQUIRE
```diff
@@ -1,5 +1,5 @@
 julia 0.5
-DiffBase 0.0.3 0.3.0
+DiffBase 0.0.3 0.4.0
```
tkelman: looks like this loosening won't matter until there's an allowed ForwardDiff version that isn't upper-bounding DiffBase to < 0.3

PR #90, opened 2017-08-28
fix #76

### LINKED ISSUE #76: In-place assignment on strictly typed vectors does not report error
If using an in-place assignment operator `.=` from a tracked vector into a typed vector, no error is reported and an incorrect answer is given. Consider the following:

```
using ReverseDiff

function test1(x)
    out = Vector{Float64}(length(x))
    out .= 3.0.*x
    out .^= 2
    return sqrt(dot(out,out))
end
function test2(x)
    out = Vector{Any}(length(x))
    out .= 3.0.*x
    out .^= 2
    return sqrt(dot(out,out))
end

input = rand(3)
test1(input) == test2(input)
g1 = ReverseDiff.gradient(test1, input)
g2 = ReverseDiff.gradient(test2, input)
```

Output in Julia 0.5:
```
julia> test1(input) == test2(input)
    true
julia> g1 = ReverseDiff.gradient(test1, input)
    3-element Array{Float64,1}:
    0.0
    0.0
    0.0
julia> g2 = ReverseDiff.gradient(test2, input)
    3-element Array{Float64,1}:
    0.0936007
    6.53833
    13.5873
```

jrevels:
I can't spend too much time on ReverseDiff maintenance at the moment, but this is basically a one-line change.

Like I said in #76, I know this definition was intentional on my part, but I can't for the life of me remember what motivated it. Let's see what tests reveal; I have a suspicion this will break a feature w.r.t. destructive and/or nested differentiation.


mohamed82008:
Closing in favor of #170.

PR #123, opened 2020-03-24
Custom gradient macro proof of concept

mohamed82008:
This PR implements a proof of concept for an `@grad` macro that enables users to define custom gradients for some functions in a user-friendly syntax, identical to Tracker's syntax. Please let me know if I made some obvious mistakes, this is my first contribution to ReverseDiff :)

Here is a usage example that is possible with this PR:
```julia
julia> using ReverseDiff, LinearAlgebra
[ Info: Precompiling ReverseDiff [37e2e3b7-166d-5795-8a7a-e32c996b4267]

julia> f(x) = dot(x, x)
f (generic function with 1 method)

julia> f(x::ReverseDiff.TrackedVector) = ReverseDiff.track(f, x)
f (generic function with 2 methods)

julia> ReverseDiff.@grad function f(x)
           println("The custom gradient has been used.")
           xv = ReverseDiff.value(x)
           return dot(xv, xv), ∇ -> (∇ * 2 * xv,)
       end

julia> x = rand(3);

julia> ReverseDiff.gradient(x -> dot(x, x), x)
3-element Array{Float64,1}:
 1.5585250070434835
 1.8206637829549495
 0.32560088933194864

julia> ReverseDiff.gradient(f, x)
The custom gradient has been used.
3-element Array{Float64,1}:
 1.5585250070434835
 1.8206637829549495
 0.32560088933194864
```

PR #125, opened 2020-04-09
fill, vcat, hcat, cat, reshape, any, and all custom adjoints

mohamed82008:
This PR defines a few custom adjoints for some array functions. Currently, it gives a precompilation error in Julia 1.3 and a StackOverFlow in Julia 1.4 for what seems like a Julia bug triggered by https://github.com/JuliaDiff/ReverseDiff.jl/blob/mt/array_funcs/src/derivatives/arrays.jl#L40 and the next block. Commenting these 2 blocks, loading ReverseDiff and then calling:
```julia
using ReverseDiff: TrackedVecOrMat, track
Base.vcat(x::TrackedVecOrMat{T}, xs::AbstractVecOrMat{T}...) where T = track(vcat, x, xs...)
```
triggers the StackOverFlow in Julia 1.4 which again seems like a Julia bug?? I appreciate any help here. This PR is based on #123.

### DIFF
--- a/src/derivatives/arrays.jl
+++ b/src/derivatives/arrays.jl
@@ -37,12 +37,15 @@ for f in [:hcat, :vcat]
         cnames = map(_ -> gensym(), c)
         @eval Base.$f($([:($x::$c) for (x, c) in zip(cnames, c)]...), x::Union{TrackedArray,TrackedReal}, xs::Union{AbstractArray,Number}...) = track($f, $(cnames...), x, xs...)
     end
+    for i = 0:2, c = combinations([:AbstractVecOrMat, :TrackedVecOrMat], i)
+        cnames = map(_ -> gensym(), c)
+        @eval Base.$f($([:($x::$c{T}) for (x, c) in zip(cnames, c)]...), x::TrackedVecOrMat{T}, xs::AbstractVecOrMat{T}...) where T = track($f, $(cnames...), x, xs...)
+    end
+    for i = 0:2, c = combinations([:AbstractVector, :TrackedVector], i)
+        cnames = map(_ -> gensym(), c)
+        @eval Base.$f($([:($x::$c{T}) for (x, c) in zip(cnames, c)]...), x::TrackedVector{T}, xs::AbstractVector{T}...) where T = track($f, $(cnames...), x, xs...)
+    end
     @eval begin
-        Base.$f(x::TrackedVecOrMat{T}, xs::AbstractVecOrMat{T}...) where T = track($f, x, xs...)
-        Base.$f(x1::TrackedVecOrMat{T}, x2::TrackedVecOrMat{T}, xs::AbstractVecOrMat{T}...) where T = track($f, x1, x2, xs...)
-        Base.$f(x::TrackedVector{T}, xs::AbstractVector{T}...) where T = track($f, x, xs...)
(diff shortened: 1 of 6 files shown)


mohamed82008:
@ChrisRackauckas any idea what's causing the stackoverflow error?

mohamed82008:
I have commented out the problematic code.

mohamed82008:
Hmm actually defining the method outside seems to work now, this is weird. I could swear it didn't earlier 😕

mohamed82008:
Ok so this seems to trigger it:
```julia
using ReverseDiff: TrackedVecOrMat, track
Base.hcat(y::AbstractVecOrMat{T}, x::TrackedVecOrMat{T}, xs::AbstractVecOrMat{T}...) where T = track(vcat, y, x, xs...)
```

mohamed82008:
So I avoided the methods that trigger the error such as the one above for now. To trigger the error, load the package and call the above method.

mohamed82008:
I will add some tests here tomorrow.

src/derivatives/arrays.jl
```diff
@@ -0,0 +1,110 @@
...
+          d
+        end
+        return (Δs...,)
+    end
+    return out_value, back
+end
+
+@grad function hcat(xs::Union{TrackedVector, TrackedMatrix}...)
+    xs_value = value.(xs)
+    out_value = hcat(xs_value...)
```
ChrisRackauckas, line 77:
```suggestion
    out_value = reduce(hcat,xs_value)
```

src/derivatives/arrays.jl
```diff
@@ -0,0 +1,110 @@
...
+        end
+        @grad function $f(x1::AbstractVector, x2::Real)
+            $f(value(x1), value(x2)), (Δ) -> (Δ[1:length(x1)], Δ[length(x1)+1])
+        end
+    end
+end
+
+@grad function vcat(xs::Union{TrackedVector, TrackedMatrix}...)
+    xs_value = value.(xs)
+    out_value = vcat(xs_value...)
```
ChrisRackauckas, line 60:
```suggestion
    out_value = reduce(vcat,xs_value)
```

mohamed82008:
I am still not too happy with this PR especially the `getindex` changes. A note to myself: I need to benchmark all the functions here against Tracker.

ChrisRackauckas:
I think it's still a lot better than before. Would you be fine with merging and continuing to improve later? I think these and the broadcast overloads https://github.com/TuringLang/DistributionsAD.jl/blob/master/src/reversediffx.jl#L558-L585 work out pretty well from my tests

mohamed82008:
I will make sure this branch's code is consistent with DistributionsAD then merge.

mohamed82008:
Will rebase and then merge once tests pass.

ChrisRackauckas:
I think it's worth adding the broadcast overloads here as well. I can see that they don't cover the general case, but it does capture a lot already.

mohamed82008:
>I can see that they don't cover the general case, but it does capture a lot already.

The "general" case is pretty hard without source-to-source. By the general case, I mean when you have intermediate arrays of structs with tracked fields. ReverseDiff cannot track structs. A struct of arrays approach might work here as a sort of lazy approach to allow fusion with future broadcasting operations. I can experiment with this when I get some time.

mohamed82008:
This PR is already big-ish. No harm in merging and creating another PR for broadcasting.

ChrisRackauckas:
agreed

PR #133, opened 2020-05-31
add an overload for zero

ChrisRackauckas:
This makes it keep TrackedArray instead changing to Array{TrackedReal}


YingboMa:
The test failure looks real.

PR #147, opened 2020-09-08
Replace TrackedReal dispatches in adjoint and transpose

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/330

PR #148, opened 2020-09-09
Fix adjoint and transpose properly

mohamed82008:
This PR fixes `Adjoint` and `Transpose` support properly without relying on dispatch workarounds. I only had an issue with `transpose` which seems to cause an unusual growth in the tape size when called. I traced the function calls and it's just calling the parametric constructor so it shouldn't be doing this. This needs  further investigation but is not a show stopper.

Supersedes: #147 
Fixes: https://github.com/SciML/DiffEqSensitivity.jl/issues/330

### DIFF
--- a/src/derivatives/linalg/arithmetic.jl
+++ b/src/derivatives/linalg/arithmetic.jl
@@ -188,23 +188,23 @@ for S1 in (:TrackedArray, :TrackedVector, :TrackedMatrix)
         @eval begin
             LinearAlgebra.:*(x::$(S1){X,D}, y::$(S2){Y,D}) where {X,Y,D} = record_mul(x, y, D)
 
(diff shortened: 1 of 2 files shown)


torfjelde:
Quick quesh: is there going to be a new release with this fix included anytime soon?:)

mohamed82008:
Yes I told @ChrisRackauckas he can do the honours once happy with it which I think he is. But it must have slipped through the cracks. I will do it now.

mohamed82008:
https://github.com/JuliaRegistries/General/pull/21421

torfjelde:
You wonderful, wonderful human being @mohamed82008 :heart: Thank you!

ChrisRackauckas:
oh sorry about that!

Issue #153, opened 2020-10-14
`StaticArray` of `TrackedReal`s doesn't currently work (and I think it could)


bzinberg:
Sorry for the lack of minimal example here, I'll add it when I get the chance.

I recently tried to differentiate through operations on `StaticArrays.SVector`s and hit two issues, both of which were easily fixable once diagnosed (unless these solutions cause other problems I'm not aware of):

1. StaticArrays defines a catch-all `+(::StaticArray, ::AbstractArray)` and ReverseDiff defines a catch-all `+(::AbstractArray, ::TrackedArray)`.   So if `+(::StaticArray, ::TrackedArray)` is called, Julia complains that the dispatch rule is ambiguous.  (Same for the other order of operands too.)  The same is the case for `-`, and I'm guessing also for `*`.

2. This line
https://github.com/JuliaDiff/ReverseDiff.jl/blob/71c5ac0cb2212f49e3b3d567b45a830614173913/src/tracked.jl#L225
assumes that `similar(t)` supports `setindex!`.  In the case of `StaticArray`s, this assumption breaks if `isbitstype(eltype(t))` is `false`, and in particular if `t` is an `StaticArray` of `TrackedReal`s.

I was able to fix these locally by:

1. Explicitly defining `+(::StaticArray, ::TrackedArray)` and `+(::TrackedArray, ::StaticArray)`, to disambiguate the dispatch rulea similarly for `-` (and probably also for `*`, though I didn't have the need for that).

2. Redefining `ReverseDiff.capture` on `StaticArray`s to do the same thing without mutation (using `map` + array conversion instead of `similar` + `map!`).  As an aside, if this is equally efficient, seems like we might as well make this the default definition.

PR #154, opened 2020-10-14
Support `StaticArray`s of `TrackedReal`s

bzinberg:
Draft fix to https://github.com/JuliaDiff/ReverseDiff.jl/issues/153.

### DIFF
--- a/src/derivatives/linalg/arithmetic.jl
+++ b/src/derivatives/linalg/arithmetic.jl
@@ -33,6 +33,9 @@ for A in ARRAY_TYPES
     @eval @inline Base.:+(x::$(A), y::TrackedArray{V,D}) where {V,D} = record_plus(x, y, D)
 end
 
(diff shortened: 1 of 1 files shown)


bzinberg:
LMK if this looks like a good approach.  If yes, then I'll add unit tests.

src/derivatives/linalg/arithmetic.jl
```diff
@@ -33,6 +33,9 @@ for A in ARRAY_TYPES
     @eval @inline Base.:+(x::$(A), y::TrackedArray{V,D}) where {V,D} = record_plus(x, y, D)
 end
 
+@inline Base.:+(x::TrackedArray{V,D}, y::StaticArray) where {V,D} = record_plus(x, Array(y), D)
```
mohamed82008, line 36: Try changing to an `MArray` instead of `Array` here.

mohamed82008: Then inplace operations will work.

mohamed82008: And please post a test of MWE that was broken.

PR #163, opened 2020-12-21
Add prevfloat and nextfloat rules

ChrisRackauckas:


src/derivatives/scalars.jl
```diff
@@ -116,3 +116,6 @@ end
     end
     return nothing
 end
+
+Base.prevfloat(r::ReverseDiff.TrackedReal) = r - eps(ReverseDiff.value(r))
```
mohamed82008, line 120: No need for `ReverseDiff.` inside `ReverseDiff` ;)

mohamed82008 (approved):

PR #164, opened 2021-01-29
Implement `view`

devmotion:
Currently, `track` + a custom gradient implementation return a gradient of zero when called with a `SubArray`. After a lot of debugging, I was able to condense one of the problems in https://github.com/TuringLang/DistributionsAD.jl/pull/151 to the following MWE:
```julia
julia> using ReverseDiff

julia> f(x::AbstractVector) = sum(abs2, x)

julia> f(x::AbstractVector{<:ReverseDiff.TrackedReal}) = ReverseDiff.track(f, x)

julia> ReverseDiff.@grad function f(x::AbstractVector{<:ReverseDiff.TrackedReal})
           xv = ReverseDiff.value(x)
           y = f(xv)
           function f_pullback(Δ::Real)
               return ((2 * Δ) .* xv,)
           end
           return y, f_pullback
       end

julia> A = rand(10, 20);

julia> ReverseDiff.gradient(A) do x
           sum(i -> sum(abs2, view(x, :, i)), axes(x, 2))
       end ≈ 2 * A
true

julia> ReverseDiff.gradient(A) do x
           sum(i -> f(x[:, i]), axes(x, 2))
       end ≈ 2 * A
true

julia> ReverseDiff.gradient(A) do x
           sum(i -> f(view(x, :, i)), axes(x, 2))
       end ≈ 2 * A
false

julia> all(
           iszero,
           ReverseDiff.gradient(A) do x
               sum(i -> f(view(x, :, i)), axes(x, 2))
           end,
       )
true
```
I.e., the gradient can be computed correctly with `view` if `track` is not used, and it can be computed correctly with the custom gradient without `view`. However, the gradients are zero if `view` is used together with a custom gradient.

This PR fixes the example above by implementing `view` for TrackedArray such that it returns a `TrackedArray` instead of a `SubArray` (as currently done). This is consistent with the definitions of `getindex` and `reshape`. However, it is a bit unfortunate that it does only work if `view` would return an array with `IndexLinear` indexing style - apparently an assertion in the implementation of `TrackedArray` only allows this indexing style.

Therefore I wonder if maybe it would be better to fix the problem by adding some special cases for `SubArray` in the `track`ing pipeline. Since I am not familiar with the details there, I chose to implement `view`. In general, `view` is not guaranteed to return a `SubArray` (e.g. `FillArray`s return `FillArray`s and `Range`s `Range`s), so special casing on `SubArray` would not fix all use cases of `view`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReverseDiff"
 uuid = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
-version = "1.5.0"
+version = "1.6.0"
 
 [deps]
 DiffResults = "163ba53b-c6d8-5494-b064-1a9d43ac40c5"
--- a/src/tracked.jl
+++ b/src/tracked.jl
@@ -385,6 +385,10 @@ reshape_body = :(TrackedArray(reshape(value(t), dims), reshape(deriv(t), dims), 
 @eval Base.reshape(t::TrackedArray, dims::Colon...) = $reshape_body
 @eval Base.reshape(t::TrackedArray, dims::Union{AbstractUnitRange,Int64,Colon}...) = $reshape_body
 
+Base.@propagate_inbounds function Base.view(t::TrackedArray, inds...)
+    return TrackedArray(view(value(t), inds...), view(deriv(t), inds...), tape(t))
+end
+
 ####################
 # `Real` Interface #
 ####################
--- a/test/MacrosTests.jl
+++ b/test/MacrosTests.jl
@@ -324,6 +324,21 @@ using ReverseDiff: @grad, TrackedReal, TrackedVector, TrackedMatrix, TrackedArra
     g2 = ReverseDiff.gradient(A -> sum(sum(A, dims = 1)), A)
     @test g1 == g2
     @test custom_grad_called
+
+    f9(x::AbstractVector) = sum(abs2, x)
+    f9(x::AbstractVector{<:TrackedReal}) = ReverseDiff.track(f9, x)
+    @grad function f9(x::AbstractVector{<:TrackedReal})
+        global custom_grad_called = true
+        xv = ReverseDiff.value(x)
+        f9(xv), Δ -> ((2 * Δ) .* xv,)
+    end
+    custom_grad_called = false
+    g1 = ReverseDiff.gradient(A) do x
+        sum(i -> f9(view(x, :, i)), axes(x, 2))
+    end
+    g2 = ReverseDiff.gradient(x -> sum(abs2, x), A)
+    @test g1 == g2
+    @test custom_grad_called
 end
 
 end # module
--- a/test/TrackedTests.jl
+++ b/test/TrackedTests.jl
@@ -1,7 +1,7 @@
 module TrackedTests
 
 using ReverseDiff, Test
-using ReverseDiff: TrackedReal, TrackedArray
+using ReverseDiff: TrackedReal, TrackedArray, TrackedVector, TrackedMatrix
 
 include(joinpath(dirname(@__FILE__), "utils.jl"))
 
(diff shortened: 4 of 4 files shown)


devmotion:
Bump 🙂

mohamed82008:
This is a breaking change because it can change dispatch behavior.

mohamed82008:
FWIW, this is not the proper way to define a custom adjoint:
```julia
f(x::AbstractVector{<:ReverseDiff.TrackedReal}) = ReverseDiff.track(f, x)
ReverseDiff.@grad function f(x::AbstractVector{<:ReverseDiff.TrackedReal})
    xv = ReverseDiff.value(x)
    y = f(xv)
    function f_pullback(Δ::Real)
        return ((2 * Δ) .* xv,)
    end
    return y, f_pullback
end
```
One should use `ReverseDiff.TrackedVector` for dispatch. Calling `track` on a vector of tracked reals is not supported and should probably throw. But changing that still doesn't make use of the custom adjoint for views. So this PR may still be useful.

mohamed82008:
If you additionally define:
```julia
f(x::SubArray{<:ReverseDiff.TrackedReal, <:Any, <:ReverseDiff.TrackedArray}) = ReverseDiff.track(f, copy(x))
```
the custom rule will work for the view of a tracked array. Note that in my experience `ReverseDiff` assumes a `Base.Array` only inside its `TrackedArray` and doesn't really handle well any other array type so I am very reluctant to approve this PR. The tests passing is a good sign but we may need to run some downstream tests as well.

devmotion:
> If you additionally define:

Sure but the main point of `view` is to avoid the copy :smile: 

Also it seems that the implementation of `track(!)` in https://github.com/JuliaDiff/ReverseDiff.jl/blob/d397cdf77c7855863c75c43bf2d3cdc7de86bfd4/src/tracked.jl#L435-L450 explicitly supports `AbstractArray{<:TrackedReal}`. Since
```julia
julia> ReverseDiff.TrackedVector{<:Real,<:Real} <: AbstractVector{<:ReverseDiff.TrackedReal}
true
```
this adds support for `TrackedVector` etc. automatically. Since `view` currently does not return a `TrackedVector` but only an `AbstractVector{<:TrackedReal}`, actually one has to use `AbstractVector{<:TrackedReal}` (or `copy`...) currently if one wants to ensure that a custom gradient is used:
```julia
julia> view(ReverseDiff.TrackedArray(rand(10), rand(10), ReverseDiff.NULL_TAPE), 1:4) isa ReverseDiff.TrackedVector
false

julia> view(ReverseDiff.TrackedArray(rand(10), rand(10), ReverseDiff.NULL_TAPE), 1:4) isa AbstractVector{<:ReverseDiff.TrackedReal}
true
```

A slightly modified version of the example above illustrates this. Let us define an incorrect adjoint here, so it's clear if the custom adjoint is used or the fallback for `sum`:
```julia
julia> f(x::AbstractVector) = sum(abs2, x)
f (generic function with 2 methods)

julia> f(x::ReverseDiff.TrackedVector) = ReverseDiff.track(f, x)
f (generic function with 2 methods)

julia> ReverseDiff.@grad function f(x::ReverseDiff.TrackedVector)
           xv = ReverseDiff.value(x)
           y = f(xv)
           function f_pullback(Δ::Real)
               return ((3 * Δ) .* xv,)
           end
           return y, f_pullback
       end

julia> A = rand(10, 20);

julia> ReverseDiff.gradient(A) do x
           sum(i -> f(x[:, i]), axes(x, 2))
       end ≈ 3 * A
true

julia> ReverseDiff.gradient(A) do x
           sum(i -> f(view(x, :, i)), axes(x, 2))
       end ≈ 3 * A
false
```
I.e., currently if a package/developer uses `view` suddenly the adjoint is not used anymore and code will break or be less efficient (which maybe is even more annoying since it is more difficult to spot).

mohamed82008:
>explicitly supports AbstractArray{<:TrackedReal}.

Sure, but it isn't supported in custom adjoints. However, adding `AbstractArray{<:TrackedReal}` to the union in https://github.com/JuliaDiff/ReverseDiff.jl/blob/71c5ac0cb2212f49e3b3d567b45a830614173913/src/macros.jl#L241 fixes that. I don't see any side-effects that this change may have. Perhaps try it?

devmotion:
It seems to work :slightly_smiling_face: I opened another PR.

mohamed82008:
Closing this since #169 was merged.

PR #166, opened 2021-02-06
Fix `Base.isapprox` for `TrackedReal`

### LINKED ISSUE #165: `isapprox` failing for `TrackedReal`
Hi all,

I'm currentlyplaying with bayesian NNs, and I'm slowly building up from the basic [example given on the Turing.jl website](https://turing.ml/dev/tutorials/3-bayesnn/), now trying to extend the classification from being binary to multiclass. Thus, I defined the output of my BNN as a `Categorical` instead of a `Binomial`. Logically enough, `Categorical` requires samples to be probability vector, positive and summing to one, _i.e._, from Distributions/src/utils.jl:

```
isprobvec(p::AbstractVector{<:Real}) =
    all(x -> x ≥ zero(x), p) && isapprox(sum(p), one(eltype(p)))
```

Naturally, output of the `Flux.softmax` of a vector is not exactly 1.0 due to floating points shenanigans, but `isapprox` takes care of that.

The problem however, is that what works with floats doesn't seem to work with `TrackedReal`:

```
xx = ReverseDiff.TrackedReal{Float64,Float64,ReverseDiff.TrackedArray{Float64,Float64,2,Array{Float64,2},Array{Float64,2}}}[TrackedReal<4WI>(0.42766067643153494, 0.0, 3Wk, ---), TrackedReal<FB4>(0.1639909617549507, 0.0, 3Wk, ---), TrackedReal<49W>(0.4083483618135143, 0.0, 3Wk, ---)]
sum(xx) = TrackedReal<KQ5>(0.9999999999999999, 0.0, 3Wk, ---)
one(eltype(xx)) = TrackedReal<HSG>(1.0, 0.0, ---, ---)
isapprox(sum(xx), 1.0) = true
isapprox(sum(xx), one(eltype(xx))) = false
isprobvec(r[:, i]) = false
```

So `xx`, the output of my network, is indeed a “very close to sum-to-one” vector of `TrackedReal` (`xx = [0.427..., 0.163..., 0.408...]`); its sum is a “very-close-to-one” `TrackedReal` (`sum = TrackedReal(0.9999999999999999, ...)`), it is virtually equal to `Float64(1.0)` (`isapprox(sum(xx), 1.0) = true`), **but**, it's not very close to the `one` of its type: `isapprox(sum(xx), one(eltype(xx))) = false`. Hence, it's not a probability vector, and it is rejected by `Categorical`.

[Here is a minimal failing program from which the above example is extracted; if you run it as is under Julia 1.5, you should encounter my problem](https://gist.github.com/delehef/2895d5e95935bf158c6221c6b9d46f10)

Did I miss something and are tracked reals failing to compare similar to one a normal behaviour, or is there a bug somewhere?

delehef:
The `Base.isapprox` function relies on `Base.rtoldefault` to return an acceptable margin of error for inexact comparisons. If the default implementation for types descending from `AbstractFloat` returns `sqrt(eps(T))`, the one for types descending from `Real` simply returns 0.

As `TrackedReal` descends from `Real` and not `AbstractFloat`, a custom `Base.rtoldefault` implementation returning `sqrt(eps(T))` ensures that `iscompare` behaves correctly for `TrackedReal` approximative comparisons.

Fixes #165

### DIFF
--- a/src/tracked.jl
+++ b/src/tracked.jl
@@ -426,6 +426,8 @@ Base.round(::Type{R}, t::TrackedReal) where {R<:Real} = round(R, value(t))
 Base.oneunit(t::TrackedReal) = one(t)
 Base.oneunit(::Type{T}) where {T<:TrackedReal} = one(T)
 
+Base.rtoldefault(::Type{T}) where {T<:TrackedReal} = sqrt(eps(T))
+
 ################
 # track/track! #
 ################
--- a/test/TrackedTests.jl
+++ b/test/TrackedTests.jl
@@ -741,6 +741,14 @@ tr_rand = rand(MersenneTwister(1), TrackedReal{Int,Float64,Nothing})
 @test round(tr_float) === round(v_float)
 @test round(Int, tr_float) === round(Int, v_float)
 
+f = rand()
+tr1 = ReverseDiff.TrackedReal(f, rand(), tp)
+tr2 = ReverseDiff.TrackedReal(2*f, rand(), tp)
+tr3 = ReverseDiff.TrackedReal(f + eps(f), rand(), tp)
+
+@test isapprox(tr1, tr2) == false
+@test isapprox(tr1, tr3) == true
+
 ################
 # track/track! #
 ################


mohamed82008:
Thanks for this and sorry for the delay. Can you please add a test?

PR #169, opened 2021-02-28
Add support for `AbstractArray{<:TrackedReal}`

devmotion:
This is an alternative to https://github.com/JuliaDiff/ReverseDiff.jl/pull/164, as suggested by @mohamed82008. It fixes the example therein by adding support for `AbstractArray{<:TrackedReal}` instead of changing the output of `view` to `TrackedArray` instead of `SubArray`. Therefore it should add support for other array wrappers as well (untested).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReverseDiff"
 uuid = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
-version = "1.5.0"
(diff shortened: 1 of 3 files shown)


mohdibntarek (approved):
Looks good, thanks for the PRs!

mohamed82008:
If tests pass, I will merge and release.

PR #173, opened 2021-04-11
Fix sum of tracked reals in Turing

mohamed82008:
This PR fixes https://github.com/TuringLang/Turing.jl/issues/1573. I still need to add a reduced test.

### DIFF
--- a/src/derivatives/linalg/reductions.jl
+++ b/src/derivatives/linalg/reductions.jl
@@ -14,6 +14,7 @@ end
 
 @noinline function special_reverse_exec!(instruction::SpecialInstruction{typeof(sum)})
     input, dims = instruction.input
+    pull_value!(input)
     output = instruction.output
     if istracked(input)
         if dims === Colon()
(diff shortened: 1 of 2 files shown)


src/derivatives/linalg/reductions.jl
```diff
@@ -5,6 +5,16 @@
...
+    tp = tape(x)
+    if dims === nothing
+        return reduce(+, x)
+    else
+        out = track(sum(value(x), dims = dims), D, tp)
+        record!(tp, SpecialInstruction, sum, (x, dims), out)
+    end
+    return out
+end
 function Base.sum(x::TrackedArray{V,D}) where {V,D}
```
devmotion, line 18: Just curious and trying to understand what's going on: could one instead of adding the definition above just widen the signature here to
```suggestion
 function Base.sum(x::AbstractArray{<:TrackedReal{V,D}}; dims=:) where {V,D}
```
and forward the keyword argument? IIRC `TrackedArray <: AbstractArray{<:TrackedReal}` so this should not break the existing definition.

mohamed82008: thanks I had to change the implementation slightly but I incorporated your suggestion

ChrisRackauckas:
Looks like this caused a downstream test failure?

https://github.com/SciML/DiffEqSensitivity.jl/pull/408

PR #180, opened 2021-07-13
[WIP] Integrate ReverseDiff with ChainRules

KDr2:


KDr2:
The next I am trying to do is to find the built-in functions that have their `rrule`s in `ChainRules` but don't have `grad` rules in `ReverseDiff.jl`, then import these `rrule`s to this package.

Does anyone have any idea about this?

KDr2:
> to find the built-in functions that have their `rrule`s in `ChainRules` but don't have `grad` rules in `ReverseDiff.jl`

Currently, I find the functions that have `rrule`s in `ChainRules` by the following code:

```jl
using ChainRulesCore, ChainRules

const noop = (x) -> x

function ruleof(method)
    parameters = if isa(method.sig, DataType)
        method.sig.parameters
    elseif method.sig.body |> typeof == DataType
        method.sig.body.parameters
    else
        []
    end
    if length(parameters) >= 2
        type2 = parameters[2]
        typeof(type2) == DataType && type2.super == Function && return type2.instance
    end

    # type2 is RuleConfig, use the next parameter
    if length(parameters) >= 3
        type3 = parameters[3]
        typeof(type3) == DataType && type3.super == Function && return type3.instance
    end

    return noop # return a function to keep type-stable
end

const methods_with_rrule = Set([ruleof(m) for m in methods(rrule)])

```
`methods_with_rrule` is a `Set{Function}`.

And use a similar way, I can find a list of functions which can be the first argument of `ReverseDiff.track`(i.e., the functions have customized derivative rules provided via `@grad`). This, plus the list from `DiffRules.diffrules()`, I get the functions who have their derivative rules in `ReverseDiff`.

So far, it seems possible to compare the two sets mentioned above, and import the rules of interest to RD. Is there any flaw in this approach?

src/macros.jl
```diff
@@ -237,6 +237,67 @@ macro grad(expr)
         end
     end |> esc
 end
+
+
+"""
+    ReverseDiff.@grad_from_cr Base.sin
```
yebai, line 243:
```suggestion
    ReverseDiff.@grad_from_chainrules Base.sin
```

test/ChainRulesTests.jl
```diff
@@ -0,0 +1,47 @@
+module ChainRulesTest
```
yebai, line 1: Maybe consider adding some tests for (subset) of functions in `FUNCS_FROM_CHAINRULES`? We need to make sure rrules for these functions are correctly imported at least.

yebai:
It seems CI is breaking for reasons not related to this PR but would be good to fix it.

test/ChainRulesTests.jl
KDr2: I found all the functions in the `Base.FastMath` (the ones I listed) already work well in ReverseDiff (without importing the `rrule`s from ChainRules), so they are not necessary to be imported. Actually, almost all the numeric computing related functions are covered by ReverseDiff. Which ChainRules has but ReverseDiff doesn't are mostly numeric-unrelated ones, so I think we can just remove the file `src/chainrules.jl` and import nothing by default. How do you think about this? @yebai

src/macros.jl
KDr2: Updated.

KDr2:
> It seems CI is breaking for reasons not related to this PR but would be good to fix it.

Yeah, and I can't reproduce it under my local ENV, and I am trying to print some debug info and see the details in the CI logs.

KDr2:
> > It seems CI is breaking for reasons not related to this PR but would be good to fix it.
> 
> Yeah, and I can't reproduce it under my local ENV, and I am trying to print some debug info and see the details in the CI logs.

https://github.com/JuliaDiff/ReverseDiff.jl/pull/180/checks?check_run_id=3357730072#step:6:288 Here is the detail. It somehow has many scalar instructions in the tape.

KDr2:
> https://github.com/JuliaDiff/ReverseDiff.jl/pull/180/checks?check_run_id=3357730072#step:6:288 Here is the detail. It somehow has many scalar instructions in the tape.

```
  testing Array -> Number functions: `#19`...
f = Main.LinAlgTests.var"#19#23"()
x = [0.8701349878780872 0.28204319051637505 0.9123350298255161; 0.006457014103689707 0.5531597766343656 0.4730983492512397; 0.773618934964177 0.08793093987132328 0.8364989250826362]
1-element InstructionTape:
tp = 1 => SpecialInstruction(*):
  input:  (1×9 adjoint(::Vector{ReverseDiff.TrackedReal{Float64, Float64, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}}}) with eltype ReverseDiff.TrackedReal{Float64, Float64, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}},
           9-element ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}})
  output: TrackedReal<E5Q>(3.5048365030607127, 0.0, KM2, ---)
  cache:  (1×9 adjoint(::Vector{Float64}) with eltype Float64,
           9-element Vector{Float64})
```
Above is the tape when run under my local ENV, it seems `adjoint` propagated to the vector's elements and the tape recorded these operations in the CI run.

test/ChainRulesTests.jl
yebai: Sounds good, but it’s probably helpful to have a placeholder for functions imported from chain rules automatically. Maybe keep one function as a minimal in the list?

test/ChainRulesTests.jl
```diff
@@ -0,0 +1,59 @@
...
+begin # test ChainRules.rrule
+    input = rand(3, 3)
+    output, back = ChainRules.rrule(f, input);
+    _, d = back(1)
+    @test output == f(input)
+    @test d == fill(3, size(input))
+end
+
+begin # test ReverseDiff
+    const f_tape = ReverseDiff.GradientTape(x -> f(x) + 2, (rand(3, 3),))
```
mohamed82008, line 39: Please test with a compiled tape as well.

src/macros.jl
```diff
@@ -237,7 +237,75 @@ macro grad(expr)
...
+                out_value = pullback(input...; kwargs...)[1]
+                $ReverseDiff.value!(output, out_value)
+                return nothing
+            end
+        end
+    end |> esc
+end
+
 _add_to_deriv!(d1, d2) = nothing
+function _add_to_deriv!(d1::Union{TrackedReal, AbstractArray{<:TrackedReal}}, t::InplaceableThunk)
```
mohamed82008, line 303: Why is this related? Is it covered in the tests?

mohamed82008:
This PR needs more tests of different types of functions, inputs, outputs, kwargs, etc. Then we need to error if the arguments are not suitable, e.g. ChainRules supports struct and arbitrary inputs but ReverseDiff doesn't the way it's defined currently. See the `@grad` macro tests for a reference.

src/macros.jl
KDr2: The derivatives returned by `ChaiRules` could be a concrete Number/Array, or a Thunk/InplaceableThunk. The latter is for lazy computing, the test case for `LinearAlgebra.norm1` tests this.

test/ChainRulesTests.jl
KDr2: updated.

KDr2:
> e.g. ChainRules supports struct and arbitrary inputs but ReverseDiff doesn't the way it's defined currently. See the `@grad` macro tests for a reference.

This PR doesn't try to introduce the diverse inputs support which `ChainRules` has, it aims at using `rrule` on the data types which `ReverseDiff` supports. Anyway, I will see the tests for `@grad` and add more for `@grad_from_chainrules`.

KDr2:
> https://github.com/JuliaDiff/ReverseDiff.jl/pull/180/checks?check_run_id=3357730072#step:6:288 Here is the detail. It somehow has many scalar instructions in the tape.

I found this is caused by a change in DiffRules (https://github.com/JuliaDiff/DiffRules.jl/commit/cbf17ea233d16deb6f0a32bca661c63a79126adc), it added `adjoint` to the diff rule list, so adjoint was recorded for each element.

And, this `adjoint` caused another test failure:
https://github.com/JuliaDiff/ReverseDiff.jl/pull/180/checks?check_run_id=3564090993#step:6:352
The gradient from ReverseDiff and ForwardDiff differ a lot, and it seems that the latter is correct.

@mohamed82008 @yebai @devmotion do you have any idea about the reason for this failure?

mohamed82008:
If you write a MWE here, I can try to look into it.

KDr2:
> If you write a MWE here, I can try to look into it.

Thank you, here is a MWE:

```julia
using ForwardDiff, ReverseDiff
using LinearAlgebra: norm

f(x) = norm(x' .* x)
x = rand(5)

test = ForwardDiff.gradient!(DiffResults.GradientResult(x), f, x)
@show DiffResults.gradient(test)

seedx = rand(eltype(x), size(x))
tp = ReverseDiff.GradientTape(f, seedx)
@show ReverseDiff.gradient!(tp, x)
```

Run it on the master branch of ReverseDiff, and ForwardDiff=0.10.18, DiffRules=1.0.2, the two results will be the same.
Run it on the master branch of ReverseDiff, and ForwardDiff=0.10.19, DiffRules=1.3.0, the two results will be different.

It is these two lines that cause the problem:
```
@define_diffrule Base.conj(x)                 = :(  1                                  )
@define_diffrule Base.adjoint(x)              = :(  1                                  )
```
https://github.com/JuliaDiff/DiffRules.jl/blob/cbf17ea233d16deb6f0a32bca661c63a79126adc/src/rules.jl#L70

KDr2:
The adjoint rule is causing other issues: 
https://github.com/JuliaDiff/ReverseDiff.jl/pull/184/checks?check_run_id=3595074063#step:6:462

Another solution is that we don't deal with `Adjoint` as before, just leave it to `ForwardDiff` (also, as an optimization), to do so, we can change
```julia
for ... in DiffRules.diffrules()
    ...
```
to something like:
```julia
rules = ...DiffRules except [adjoint, conj]
for ... in rules
    ...
```
to fix the problem.

What do think @yebai @mohamed82008 @devmotion

src/macros.jl
```diff
@@ -237,7 +237,75 @@ macro grad(expr)
...
+"""
+    ReverseDiff.@grad_from_chainrules Base.sin
+
+The `@grad_from_chainrules` macro provides a way to import adjoints defined
+in ChainRules to ReverseDiff.
+"""
+macro grad_from_chainrules(f)
+    @gensym tp output_value output back closure cls_args cls_kwargs
+    return quote
+        $f(args::Vararg{Union{ReverseDiff.TrackedReal, ReverseDiff.TrackedArray}}) = ReverseDiff.track($f, args...)
```
devmotion, line 251: I'm not familiar with all internals of ReverseDiff but this seems wrong and I don't think it will work in general. The signature is too wide if the function can only work with numbers and it does not allow to take derivatives with respect to only some of the arguments (this would require definitions for mixed tracked and non-tracked types similar to https://github.com/JuliaDiff/ReverseDiff.jl/blob/01041c8e8237ed42f6414c6fe0f6e6b12162b6ac/src/derivatives/scalars.jl#L12).

BTW it will also define a zero-argument version of the function if I'm not mistaken.

KDr2: > signature is too wide if the function can only work with numbers and it does not allow to take derivatives with respect to only some of the arguments

So, If I understand correctly, I should find all the method definitions of function f, define at least one method for each method I find, only changing the arguments of Real/Array types to its tracked version.

for example, to
```
f(::String, ::Real, ::Array{Real})
```
there should be:

```
f(::String, ::TrackedReal, ::TrackedArray)
f(::String, ::TrackedReal, ::Array)
f(::String, ::Real, ::TrackedArray)
```
three overloads, right?

devmotion: Yes, in this example you would have to define 3 methods. In general, you can't just replace `Real` and `Array{<:Real}`. Eg, scalar rules defined with `@scalar_rule` are defined for `Number` arguments by default. And arrays are also tricky. Eg, there could be two rules for a function with an `AbstractArray{<:Number}` argument where one of them is specialized for `Diagonal`. Then you want to make sure that tracked diagonal matrices end up in the specialized `rrule`. And similarly, if there is only an `rrule` for `Diagonal` you don't want to add a ReverseDiff rule for general tracked arrays.

KDr2: So we can let the user provide a method signature to import the method's rrule, instead of importing all methods of a function by providing only the function name, like:

```julia
@grad_from_chainrules f(::String, ::Number)
```

KDr2: So we can let the user provide a method signature to import the method's rrule, instead of importing all methods of a function by providing only the function name, like:

```julia
@grad_from_chainrules f(::String, ::Number)
```

devmotion: This could be done. There are still problems though for multiple arguments and due to limitations of the pattern matching, in particular for array arguments and if the rule is not defined for `Number` but some (abstract or concrete) subtype. I think a cleaner and safer approach would be to require to specify the exact combination of ReverseDiff types, i.e., e.g.
```julia
@grad_from_chainrules f(::String, ::TrackedReal)
```
and specify the desired combinations of `TrackedReal` and `TrackedArray` etc. manually.

KDr2:
Oh, `ChainRulesCore` had had many updates lately while `ChainRules` hadn't, this makes many `rrule`s don't work any more, for example:

```julia
using LinearAlgebra, ChainRules
input = rand(3, 3)
output, back = ChainRules.rrule(LinearAlgebra.norm1, input);
_, d = back(1) 
```
throws an exception on the latest `ChainRules` and `ChainRulesCore`.

devmotion:
It works fine for me with the latest versions for ChainRules and ChainRulesCore.

KDr2:
Sorry, you are right. I used a local `ChainRules` repo which was checked out on an old tag, and I got myself under an impression that it is not updated lately by running `git log` in the repo.
The test errors are still the `adjoint` related ones.
 Greetings.

Killian Zhuo (KDr2, https://kdr2.com)

 

    On Friday, September 17, 2021, 04:49:57 PM GMT+8, David Widmann ***@***.***> wrote:  
 
 


It works fine for me with the latest versions for ChainRules and ChainRulesCore.

Project.toml
```diff
@@ -3,6 +3,7 @@ uuid = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
 version = "1.9.0"
 
 [deps]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
```
devmotion, line 6: Can you add a compat bound?

src/macros.jl
```diff
@@ -237,7 +237,126 @@ macro grad(expr)
         end
     end |> esc
 end
+
+
+function funcall_fwd_to_rule(expr)
```
devmotion, line 242: Can you add some explanations and comments to the code?

src/macros.jl
```diff
@@ -237,7 +237,126 @@ macro grad(expr)
...
+    expr.head == :call || error("The rule should be in a format of a function call.")
+    func = expr.args[1]
+    has_tracked_data = false
+
+    args_l = Any[func]
+    args_r = Any[:(ReverseDiff.track)]
+    args_start = 2
+    if (expr.args[2].head == :parameters) # has kw args
```
devmotion, line 250:
```suggestion
    if expr.args[2].head == :parameters # has kw args
```

src/macros.jl
```diff
@@ -237,7 +237,126 @@ macro grad(expr)
...
+        args_start = 3
+    end
+    push!(args_r, func)
+    for arg in expr.args[args_start:end]
+        if arg.head == :(::)
+            arg_name = gensym(:arg)
+            arg_ex = :($arg_name::$(arg.args[end]))
+            push!(args_l, arg_ex)
+            push!(args_r, arg_ex)
+            if eval(arg.args[end]) <: TrackedReal || eval(arg.args[end]) <: TrackedArray
```
devmotion, line 262: Can we avoid `eval`?

src/macros.jl
```diff
@@ -237,7 +237,126 @@ macro grad(expr)
...
+                output, input = instruction.output, instruction.input
+                $ReverseDiff.pull_value!.(input)
+                pullback = instruction.cache[2]
+                kwargs = instruction.cache[3]
+                out_value = pullback(input...; kwargs...)[1]
+                $ReverseDiff.value!(output, out_value)
+                return nothing
+            end
+        end
+    end |> esc
```
devmotion, line 350: Can we not escape the whole expression but only the parts that have to be escaped (like the function name and user-provided types) and let macro hygiene take care of the rest? Similar to e.g. the macros in ChainRulesCore such as https://github.com/JuliaDiff/ChainRulesCore.jl/blob/47389f5d4748209a88c5eb72a7cdf50cdd8bfa2d/src/rule_definition_tools.jl#L246-L254.

src/macros.jl
```diff
@@ -237,7 +237,126 @@ macro grad(expr)
...
+end
+
 _add_to_deriv!(d1, d2) = nothing
+function _add_to_deriv!(d1::Union{TrackedReal, AbstractArray{<:TrackedReal}}, t::InplaceableThunk)
+    _add_to_deriv!(d1, t.val)
+end
+function _add_to_deriv!(d1::Union{TrackedReal, AbstractArray{<:TrackedReal}}, t::Thunk)
+    increment_deriv!(d1, unthunk(t))
+end
```
devmotion, line 359: Maybe a more orthogonal design would be better, i.e., maybe we could just use `unthunk` in the calls to `_add_to_deriv!` above instead of adding more dispatches. E.g,
```julia
_add_to_deriv!.(input, unthunk.(input_derivs))
```
instead of  `_add_to_deriv!.(input, input_derivs)`. Alternatively, if something more general is needed, we could define a function such as
```julia
unwrap_chainrules_output(x) = x
unwrap_chainrules_output(x::AbstractThunk) = unthunk(x)
...
```
and use it in the call to `_add_to_deriv!`.

In any case, we should not specialize on `Thunk` or `InplaceableThunk` but on `AbstractThunk`, and not access internal fields but just use `unthunk`.

test/ChainRulesTests.jl
```diff
@@ -0,0 +1,75 @@
...
+        the importing mechanism works.
+        =#
+        return ChainRules.NoTangent(), fill(3 * d, size(x))
+    end
+    return r, back
+end
+
+ReverseDiff.@grad_from_chainrules f(::ReverseDiff.TrackedArray)
+
+begin # test ChainRules.rrule
```
devmotion, line 30: `begin` could/should be removed and possibly be replaced with `@testset`. It does not introduce a new scope, if this was the intention.

test/ChainRulesTests.jl
```diff
@@ -0,0 +1,75 @@
...
+        After importing this rrule into ReverseDiff, if we get 3
+        rather than 4 when we compute the derivative of `f`, it means
+        the importing mechanism works.
+        =#
+        return ChainRules.NoTangent(), fill(3 * d, size(x))
+    end
+    return r, back
+end
+
+ReverseDiff.@grad_from_chainrules f(::ReverseDiff.TrackedArray)
```
devmotion, line 28: We need more tests, in particular with mixed and also untracked types.

Project.toml
KDr2: updated.

src/macros.jl
KDr2: updated.

src/macros.jl
KDr2: updated.

src/macros.jl
KDr2: updated.

test/ChainRulesTests.jl
KDr2: updated.

src/macros.jl
KDr2: There are too many symbols to preserve(`esc`) in this piece of code, using `gensym` to generate new symbols and keep the macro hygiene in this way makes it more readable, I think it was also because of this, the macro `@grad` was written in this style?

src/macros.jl
KDr2: OK, I changed it to using an exhaustive list of tracked types to predicate the argument types now.

src/macros.jl
```diff
@@ -237,7 +237,135 @@ macro grad(expr)
...
+        else
+            push!(args_l, arg)
+            push!(args_r, arg)
+        end
+    end
+
+    has_tracked_data || error("The rule should have at least one tracked argument.")
+    left = Expr(:call, args_l...)
+    right = Expr(:call, args_r...)
+    return expr.args[1], :($left = $right)
```
yebai, line 286:
```suggestion
    return func, :($left = $right)
```

yebai (approved):

src/macros.jl
```diff
@@ -237,7 +237,128 @@ macro grad(expr)
...
+    fcall.head == :call || error("The rule should be in format of a function call.")
+    @capture(fcall, f_(xs__)) # extract information into f and xs
+    xs_l, xs_r = _make_fwd_args(f, xs)
+
+    return quote
+        $f($(xs_l...)) = $ReverseDiff.track($(xs_r...))
+        if !hasmethod($ReverseDiff.track, Tuple{typeof($f)})
+            function $ReverseDiff.track(::typeof($f), args...; kwargs...)
+                $tp = $ReverseDiff.tape(args...)
+                $output_value, $back = ChainRules.rrule($f, map(ReverseDiff.value, args)...; kwargs...)
```
mohamed82008, line 310: can we $ ChainRules here and in other places and avoid depending on CR or CRC entirely?

src/ReverseDiff.jl
```diff
@@ -17,6 +17,8 @@ using StaticArrays
 
 using MacroTools
 
+using ChainRulesCore
```
mohamed82008, line 20: do we need to depend on this?

mohamed82008: the implementation here allows RD rules to be defined using ChainRules. It doesn't require RD to depend or have access to `rrule` as long as the user's scope has ChainRules loaded.

src/macros.jl
```diff
@@ -237,7 +237,128 @@ macro grad(expr)
...
+adjoints(rrule) defined in ChainRules to ReverseDiff. One must provide
+a method signature to import the corresponding `rrule`. In the
+provided method signature, one should replace the types of arguments
+to which one wants to take derivatives with respect with
+`ReverseDiff.TrackedReal` and `ReverseDiff.TrackedArray`
+respectively. For example, we can import `rrule` of `f(x::Real,
+y::Array)` like below:
+
+
+```
```
oxinabox, line 291: Do you need to mark code as julia for Documenter to highlight it?
````suggestion
```julia
````

Or does it default?

File: src/ReverseDiff.jl (hunk 16 above)
Comment (oxinabox) [reply src/ReverseDiff.jl:20]: Right.
You *do* need ChainRulesCore in your dependencies, since you need to lock to a particular set of versions of the API
And you already have it loaded from your downstream, so this line will be a no-op.

Also if you wanted you can fix it to not depend on ChainRulesCore being called ChainRulesCore in the users scope.
Or even loaded there.
IIRC you can do that by interpilating the module object into the expression?

**Comment by oxinabox** [comment]
I think this wants tests that use some of ChainRule's more advanced features.
Like `InplacableThunks` (which probably just work, though you would want to make some more changes to take advantage of them)
and things that return `Tangent{T}` like the `getproperty` rule for cholesky
(which probably won't work, and should be documented)

File: src/ReverseDiff.jl (hunk 16 above)
Comment (mohamed82008) [reply src/ReverseDiff.jl:20]: yes if we depend on CRC we might as well interpolate it in the macro

File: src/macros.jl (hunk 10 above)
Comment (devmotion) [reply src/macros.jl:355]: I just think nobody cared when the other macro was written and/or nobody wanted to deal with macro hygiene more carefully. Also initially the helpers in ChainRulesCore just escaped everything and were only changed a bit later.

Without `esc` you could get rid of most of the interpolations (`$ReverseDiff` etc.), which would even improve readability it seems. We mainly have to preserve (`esc`) `f` and the types in the user-provided function signature (which we have to handle anyway in `_make_fwd_args`). We could even avoid escaping the gensymed variables (and remove them completely) if we use custom argument names such as `args1` etc in the function signature (so that it is ensure there are no clashes with variable names in the function body).

File: src/macros.jl (hunk 15 above)
Comment (devmotion) [inline src/macros.jl:310]: This assumes that users load ChainRules. As suggested above, we should interpolate ChainRulesCore (similar to `$ReverseDiff`) or just use `rrule` and not escape the whole expression.

File: src/macros.jl [hunk 18]
```diff
@@ -237,7 +237,128 @@ macro grad(expr)
...
+    @gensym tp output_value output back closure cls_args cls_kwargs
+
+    fcall.head == :call || error("The rule should be in format of a function call.")
+    @capture(fcall, f_(xs__)) # extract information into f and xs
+    xs_l, xs_r = _make_fwd_args(f, xs)
+
+    return quote
+        $f($(xs_l...)) = $ReverseDiff.track($(xs_r...))
+        if !hasmethod($ReverseDiff.track, Tuple{typeof($f)})
+            function $ReverseDiff.track(::typeof($f), args...; kwargs...)
```
devmotion, line 308: Shouldn't we only define `track` for arguments and types in `xs_l...` instead of general `args...`?

test/ChainRulesTests.jl
```diff
@@ -0,0 +1,137 @@
+module ChainRulesTest
+
+using LinearAlgebra
+using ChainRules
```
devmotion, line 4: This hides the problem described above. Maybe it would be good to not load ChainRules but only ChainRulesCore?

test/ChainRulesTests.jl
```diff
@@ -0,0 +1,137 @@
...
+    function back(d)
+        # same as above, use 3 and 5 as the derivatives
+        return ChainRules.NoTangent(), fill(3 * d, size(x)), fill(5 * d, size(x))
+    end
+    return r, back
+end
+
+ReverseDiff.@grad_from_chainrules g(x::ReverseDiff.TrackedArray, y)
+ReverseDiff.@grad_from_chainrules g(x, y::ReverseDiff.TrackedArray)
+ReverseDiff.@grad_from_chainrules g(x::ReverseDiff.TrackedArray, y::ReverseDiff.TrackedArray)
```
devmotion, line 44: It would be good to test that actually after the first `@grad_from_chainrules` call only the desired argument are tracked etc.

Issue #183, opened 2021-09-14
Incorrect result on function contains adjoint.


KDr2:
As mentioned at https://github.com/JuliaDiff/ReverseDiff.jl/pull/180#issuecomment-916868570, when a function contains a call to ``adjoint`, ReverseDiff (with latest dependencies) gives a gradient that differs from other AD packages:

This is an MWE:
```julia
using ForwardDiff, ReverseDiff
using LinearAlgebra: norm

f(x) = sum(x' .* x)
x = rand(5)

test = ForwardDiff.gradient!(DiffResults.GradientResult(x), f, x)
@show DiffResults.gradient(test)

seedx = rand(eltype(x), size(x))
tp = ReverseDiff.GradientTape(f, seedx)
@show ReverseDiff.gradient!(tp, x)
```
Run it on the master branch of ReverseDiff, and ForwardDiff=0.10.18, DiffRules=1.0.2, the two results will be the same.
Run it on the master branch of ReverseDiff, and ForwardDiff=0.10.19, DiffRules=1.3.0, the two results will be different.

The problem was found after DiffRules added these two rules lately:
```julia
@define_diffrule Base.conj(x)                 = :(  1                                  )
@define_diffrule Base.adjoint(x)              = :(  1                                  )
```

https://github.com/JuliaDiff/DiffRules.jl/blob/cbf17ea233d16deb6f0a32bca661c63a79126adc/src/rules.jl#L70

devmotion:
Note that it's not a general problem with `adjoint`, eg it works fine without broadcasting:
```julia
julia> using ForwardDiff, ReverseDiff

julia> f(x) = sum(x' * x);

julia> x = rand(5);

julia> ForwardDiff.gradient(f, x) ≈ 2 * x
true

julia> ReverseDiff.gradient(f, x) ≈ 2 * x
true
```
whereas
```julia
julia> f(x) = sum(x' .* x);

julia> ForwardDiff.gradient(f, x) ≈ fill(2 * sum(x), length(x))
true

julia> ReverseDiff.gradient(f, x) ≈ fill(2 * sum(x), length(x))
false

# but:
julia> ReverseDiff.gradient(f, x) ≈ fill(sum(x), length(x))
true
```

So I think we should not opt out of the `adjoint` rules in DiffRules but use these optimizations and only fix the broken cases with broadcasting.

devmotion:
[...]

PR #185, opened 2021-09-16
Fix transposes and adjoints

devmotion:
Try to fix https://github.com/JuliaDiff/ReverseDiff.jl/issues/183, as suggested in https://github.com/JuliaDiff/ReverseDiff.jl/issues/183#issuecomment-921287528.

### DIFF
--- a/src/derivatives/arrays.jl
+++ b/src/derivatives/arrays.jl
@@ -121,3 +121,31 @@ end
         return (Δs...,)
     end
 end
+
+#########################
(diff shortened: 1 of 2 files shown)


devmotion:
The adjoint and transpose issues are fixed but now I also get the `lu` error :smile: It seems it requires a different/additional fix.

devmotion:
I think there's a better solution: https://github.com/JuliaDiff/ReverseDiff.jl/issues/183#issuecomment-921514778

PR #191, opened 2021-11-29
Fix deprecations in DiffRules 1.4

devmotion:
This PR fixes the deprecations introduced by DiffRules 1.4, and in particular the unsafe use of `DiffRules.diffrules()`.

I used the opportunity to load the rules for LogExpFunctions from DiffRules as well (which were added in DiffRules 1.4 and revealed that most downstream packages did not check if functions for which they want to define rules are actually available).


src/derivatives/broadcast.jl
```diff
@@ -264,8 +264,10 @@ end
 
 @inline _materialize(f, args) = broadcast(f, args...)
 
-for (M, f, arity) in DiffRules.diffrules()
-    isdefined(ReverseDiff, M) || continue
+for (M, f, arity) in DiffRules.diffrules(; filter_modules=nothing)
+    if !(isdefined(@__MODULE__, M) && isdefined(getfield(@__MODULE__, M), f))
```
mohamed82008, line 268: Do we know which functions will not be defined here?

devmotion: None currently. The problem is that this check was missing and therefore introducing new rules (such as the ones for LogExpFunctions) broke ReverseDiff (and ForwardDiff and Tracker...). Therefore we had to add the `filter_modules` keyword argument and set its default value to the list of the previous list of packages (i.e., SpecialFunctions and NaNMath). This is the fix for the initial problem and will avoid such issues in the future, if rules for other packages are added to DiffRules.

mohamed82008: I see. It might be useful to know if there are rules added to DiffRules and missed by ReverseDiff, so maybe turn this to a warning instead?

devmotion: OK, I added a warning in the package and an error in the tests (to ensure that all available rules are tested but to avoid breaking the package if DiffRules supports additional packages that are not loaded).

test/derivatives/ScalarTests.jl
```diff
@@ -133,15 +141,25 @@ function test_skip(f, a, b, tp)
     @test isempty(tp)
 end
 
-DOMAIN_ERR_FUNCS = (:asec, :acsc, :asecd, :acscd, :acoth, :acosh)
+# ensure that input is in domain of function
+function modify_input(f, x)
```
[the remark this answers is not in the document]
devmotion: Done.

devmotion:
@mohamed82008 It should be ready now.

mohamed82008 (approved):
LGTM.

mohamed82008:
Thanks David!

PR #192, opened 2021-12-14
Fix broken tests

devmotion:
https://github.com/JuliaDiff/DiffRules.jl/pull/73 broke the tests since, similar to `polygamma` the current DiffRules integration and test infrastructure can't deal with functions with two arguments of which one has to be an integer.

Ideally, this would be tested properly but this PR is a temporary solution that requires minimal changes and fixes also the downstream tests in DiffRules.

It seems one problem with these functions is that the arguments are promoted when they are forwarded to ForwardDiff since everything is packed in a static vector - in these cases it would be better to not promote and work with the DiffRules definitions directly. I wonder why this is not done already and instead ReverseDiff calls ForwardDiff?


ChrisRackauckas:
> It seems one problem with these functions is that the arguments are promoted when they are forwarded to ForwardDiff since everything is packed in a static vector - in these cases it would be better to not promote and work with the DiffRules definitions directly. I wonder why this is not done already and instead ReverseDiff calls ForwardDiff?

Yeah that's odd and I don't know why, but I also know this is the right solution 😅

PR #198, opened 2022-04-08
Add overloads for fld, cld, and div

sethaxen:
This PR adds overloads for `fld`, `cld`, and `div`, which all return a `MethodError` on master. The overloads for `div` are the same ones used by ForwardDiff: https://github.com/JuliaDiff/ForwardDiff.jl/blob/29332b12075e66e056843d3b4207d856f98772b3/src/dual.jl#L287-L291


oxinabox (approved):
LGTM

Issue #139, opened 2020-07-21
Problem with getindex for ReverseDiff.TrackedArray if Integer is a UInt16


mjohnson541:
Started getting this error after updating from 1.2.0 to 1.4.1:  

MethodError: no method matching ReverseDiff.TrackedArray(::Float64, ::Float64, ::Array{ReverseDiff.AbstractInstruction,1})
Closest candidates are:
  ReverseDiff.TrackedArray(!Matched::AbstractArray{V,N}, !Matched::AbstractArray{D,N}, ::Array{ReverseDiff.AbstractInstruction,1}) where {V, D, N} at ~/.julia/packages/ReverseDiff/tCOFA/src/tracked.jl:86

Stacktrace:
 [1] getindex(::ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}}, ::UInt16) at ~/.julia/packages/ReverseDiff/tCOFA/src/tracked.jl:337

When the indexes are changed to Int64 no error is thrown.

amartinhuertas:
Hi everyone, 

does anyone know if this issue was ever addressed? 

Thanks!
@amartinhuertas

kishore-nori:
We encounter a similar error when the there there is a `getindex` call on `ReverseDiff.TrackedArray` using `Int32`. Following is the error:
```julia
ERROR: MethodError: no method matching ReverseDiff.TrackedArray(::Float64, ::Float64, ::Vector{ReverseDiff.AbstractInstruction})
Closest candidates are:
 ReverseDiff.TrackedArray(::AbstractArray{V, N}, ::AbstractArray{D, N}, ::Vector{ReverseDiff.AbstractInstruction}) where {V, D, N} at ~/.julia/packages/ReverseDiff/c21a6/src/tracked.jl:83
Stacktrace:
  [1] getindex(t::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}, _inds::Int32)
```

Will try to come up with an MWE!

PR #201, opened 2022-05-25
fix for issue #139 - to allow other Int subtypes to this getindex

kishore-nori:
Fix for the following error

```julia 
ERROR: MethodError: no method matching ReverseDiff.TrackedArray(::Float64, ::Float64, ::Vector{ReverseDiff.AbstractInstruction})
Closest candidates are:
 ReverseDiff.TrackedArray(::AbstractArray{V, N}, ::AbstractArray{D, N}, ::Vector{ReverseDiff.AbstractInstruction}) where {V, D, N} at ~/.julia/packages/ReverseDiff/c21a6/src/tracked.jl:83
Stacktrace:
  [1] getindex(t::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}, _inds::Int32)
```

so that it accesses `Base.getindex(t::TrackedArray, i::Int) = TrackedReal(value(t)[i], deriv(t)[i], tape(t), i, t)` instead of `Base.getindex(t::TrackedArray, _inds::Union{Integer, Colon, AbstractArray{<:Integer}}...)`. `Base.getindex(t::TrackedArray, i::Int) = TrackedReal(value(t)[i], deriv(t)[i], tape(t), i, t)` resolves the error!

I could also change it from `Integer` to `Union{Int32,Int64}` if some thing else breaks in relation to this

### DIFF
--- a/src/tracked.jl
+++ b/src/tracked.jl
@@ -280,7 +280,7 @@ Base.promote_rule(::Type{TrackedReal{V1,D1,O1}}, ::Type{TrackedReal{V2,D2,O2}}) 
 # AbstractArray Interface #
 ###########################
 
-Base.getindex(t::TrackedArray, i::Int) = TrackedReal(value(t)[i], deriv(t)[i], tape(t), i, t)
+Base.getindex(t::TrackedArray, i::Integer) = TrackedReal(value(t)[i], deriv(t)[i], tape(t), i, t)
 
 colon2range(s, i) = i
 colon2range(s, ::Colon) = s


devmotion:
It seems that fix is not sufficient. With the PR I get
```julia
julia> x = ReverseDiff.track(rand(3));

julia> x[1]
TrackedReal<6TI>(0.09345719861940405, 0.0, Ceh, 1, AYp)

julia> x[UInt(1)]
ERROR: MethodError: no method matching ReverseDiff.TrackedReal(::Float64, ::Float64, ::Vector{ReverseDiff.AbstractInstruction}, ::UInt64, ::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}})
Closest candidates are:
  ReverseDiff.TrackedReal(::V, ::D, ::Vector{ReverseDiff.AbstractInstruction}, ::Int64, ::O) where {V, D, O} at ~/.julia/dev/ReverseDiff/src/tracked.jl:59
  ReverseDiff.TrackedReal(::V, ::D, ::Vector{ReverseDiff.AbstractInstruction}) where {V, D} at ~/.julia/dev/ReverseDiff/src/tracked.jl:61
  ReverseDiff.TrackedReal(::V, ::D) where {V, D} at ~/.julia/dev/ReverseDiff/src/tracked.jl:61
  ...
Stacktrace:
 [1] getindex(t::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}, i::UInt64)
   @ ReverseDiff ~/.julia/dev/ReverseDiff/src/tracked.jl:283
 [2] top-level scope
   @ REPL[40]:1
```

kishore-nori:
I think I focused only on including Int32 (which was source of error my case), thank you very much for adding a general fix!

PR #202, opened 2022-05-25
Fix `getindex`

### LINKED ISSUE #139: Problem with getindex for ReverseDiff.TrackedArray if Integer is a UInt16

devmotion:
This PR fixes `getindex` for non-`Int` indices (fixes #139), removes undesired definitions for `getindex` caused by Varargs, adds `@propagate_inbounds` annotations, and adds tests.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReverseDiff"
 uuid = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
-version = "1.14.0"
+version = "1.14.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


devmotion:
As a general observation, I think it would be nice if it would be possible to e.g. rely on `to_index`/`to_indices` and use a more orthogonal design, i.e., without specifying `getindex` for all kinds of indices.

yebai (approved):

PR #203, opened 2022-06-09
promotion rules for `Irrational` and `TrackedReal`

Moelf:
fix #100

PR #207, opened 2022-09-24
Fix and test `promote_rule` definitions

### LINKED ISSUE #206: Missing Promotion Rule
https://buildkite.com/julialang/scimlsensitivity-dot-jl/builds/1709#0183703e-7b3a-409e-a85b-1a59fc1a2158/832-859

```
<html>
<body>
<!--StartFragment-->

│   value =
--
  | │    MethodError: promote_rule(::Type{ForwardDiff.Dual{ForwardDiff.Tag{DiffEqBase.OrdinaryDiffEqTag, ReverseDiff.TrackedReal{Float64, Float64, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}}}, ReverseDiff.TrackedReal{Float64, Float64, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}}, 1}}, ::Type{ReverseDiff.TrackedReal{Float64, Float64, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}}}) is ambiguous. Candidates:
  | │      promote_rule(::Type{R}, ::Type{ReverseDiff.TrackedReal{V, D, O}}) where {R<:Real, V, D, O} in ReverseDiff at /root/.cache/julia-buildkite-plugin/depots/ea52448d-f230-4619-b27a-2d98107bd215/packages/ReverseDiff/5MMPp/src/tracked.jl:276
  | │      promote_rule(::Type{ForwardDiff.Dual{T, V, N}}, ::Type{R}) where {T, V, N, R<:Real} in ForwardDiff at /root/.cache/julia-buildkite-plugin/depots/ea52448d-f230-4619-b27a-2d98107bd215/packages/ForwardDiff/pDtsf/src/dual.jl:426
  | │    Possible fix, define
  | │      promote_rule(::Type{ForwardDiff.Dual{T, V, N}}, ::Type{ReverseDiff.TrackedReal{V, D, O}}) where {T, V, N, V, D, O}

<!--EndFragment-->
</body>
</html>
```

devmotion:
This PR fixes the `promote_rule` definitions by using basically the same implementation as ForwardDiff (https://github.com/JuliaDiff/ForwardDiff.jl/blob/6a6443b754b0fcfb4d671c9a3d01776df801f498/src/dual.jl#L440-L452).

Currently, abstract types (such as `Real`, `AbstractFloat` etc.) are not handled properly as `promote_rule` should be defined for `Type{<:T}` for these types instead of `Type{T}`. Moreover, currently only one order of arguments is defined which can lead to issues (usually stackoverflow errors when calling `promote_type`: https://github.com/JuliaLang/julia/issues/13193) if some other package (such as ForwardDiff) defines a `promote_rule` with reversed order of arguments and different result.

Fixes #206. Closes https://github.com/JuliaDiff/ReverseDiff.jl/pull/203.


src/ReverseDiff.jl
```diff
@@ -24,7 +24,7 @@ using ChainRulesCore
 # Not all operations will be valid over all of these types, but that's okay; such cases
 # will simply error when they hit the original operation in the overloaded definition.
 const ARRAY_TYPES = (:AbstractArray, :AbstractVector, :AbstractMatrix, :Array, :Vector, :Matrix)
-const REAL_TYPES = (:Bool, :Integer, :(Irrational{:e}), :(Irrational{:π}), :Rational, :BigFloat, :BigInt, :AbstractFloat, :Real, :Dual)
+const REAL_TYPES = (:Bool, :Integer, :(Irrational{:ℯ}), :(Irrational{:π}), :Rational, :BigFloat, :BigInt, :AbstractFloat, :Real, :Dual)
```
devmotion, line 27: This was a bug it seems:
```julia
julia> typeof(ℯ)
Irrational{:ℯ}

julia> typeof(MathConstants.e)
Irrational{:ℯ}
```

src/tracked.jl
```diff
@@ -270,10 +270,22 @@ Base.convert(::Type{T}, t::T) where {T<:TrackedReal} = t
 Base.convert(::Type{T}, t::T) where {T<:TrackedArray} = t
 
 for R in REAL_TYPES
-    @eval Base.promote_rule(::Type{$R}, ::Type{TrackedReal{V,D,O}}) where {V,D,O} = TrackedReal{promote_type($R,V),D,O}
+    R === :Dual && continue # ForwardDiff.Dual is handled below
```
devmotion, line 273: Only defining `promote_rule` for `Type{<:Dual}` leads to method ambiguity issues; hence it is skipped here and defined below for `Type{Dual{T,V,N}}`.

src/tracked.jl
```diff
@@ -270,10 +270,22 @@ Base.convert(::Type{T}, t::T) where {T<:TrackedReal} = t
...
+            Base.promote_rule(::Type{TrackedReal{V,D,O}}, ::Type{$R}) where {V,D,O} = TrackedReal{promote_type(V,$R),D,O}
+            Base.promote_rule(::Type{$R}, ::Type{TrackedReal{V,D,O}}) where {V,D,O} = TrackedReal{promote_type($R,V),D,O}
+        else
+            Base.promote_rule(::Type{TrackedReal{V,D,O}}, ::Type{R}) where {V,D,O,R<:$R} = TrackedReal{promote_type(V,R),D,O}
+            Base.promote_rule(::Type{R}, ::Type{TrackedReal{V,D,O}}) where {R<:$R,V,D,O,} = TrackedReal{promote_type(R,V),D,O}
+        end
+    end
 end
 
-Base.promote_rule(::Type{R}, ::Type{TrackedReal{V,D,O}}) where {R<:Real,V,D,O} = TrackedReal{promote_type(R,V),D,O}
```
devmotion, line 276: This is covered by the loop above (`:Real` is part of `REAL_TYPES`).

ChrisRackauckas (approved):

devmotion:
It seems DiffTests 0.1.2 (recently released) broke the ReverseDiff tests :cry:

src/tracked.jl
```diff
@@ -270,10 +270,22 @@ Base.convert(::Type{T}, t::T) where {T<:TrackedReal} = t
 Base.convert(::Type{T}, t::T) where {T<:TrackedArray} = t
 
 for R in REAL_TYPES
-    @eval Base.promote_rule(::Type{$R}, ::Type{TrackedReal{V,D,O}}) where {V,D,O} = TrackedReal{promote_type($R,V),D,O}
+    R === :Dual && continue # ForwardDiff.Dual is handled below 
+    @eval begin
+        if isconcretetype($R) # issue ForwardDiff#322
+            Base.promote_rule(::Type{TrackedReal{V,D,O}}, ::Type{$R}) where {V,D,O} = TrackedReal{promote_type(V,$R),D,O}
```
mohamed82008, line 276: Since we are implementing both methods, do we know if they are all covered by tests? Might be worth adding a test checking that switching arguments gives the same result.

src/ReverseDiff.jl
mohamed82008: Was it never tested? Would be nice to add a test.

mohamed82008:
I left some comments. It would also be nice to run some downstream tests to make sure this doesn't break anything. DiffEqSensitivity, DistributionsAD and GalacticOptim would be good to test.

src/tracked.jl
devmotion: Both are tested since `promote_type` calls both `promote_rule` definitions.

src/ReverseDiff.jl
devmotion: I already added a test (I only noticed the bug since it failed without the fix).

src/tracked.jl
devmotion: I should add that actually tests fail when one of the definitions is removed due to eg stackoverflow errors when calling `promote_type` as mentioned in the OP.

devmotion:
> DiffEqSensitivity, DistributionsAD and GalacticOptim would be good to test.

I ran tests of SciMLSensitivity, DistributionsAD (with `ENV["AD"] = "ReverseDiff"), and Optimization locally. Optimization and DistributionsAD succeeded, SciMLSensitivity tests did not finish before my laptop ran out of battery. I can retry tomorrow.

devmotion:
Took forever but tests passed with this PR in the end:
```julia
10702.904078 seconds (22.35 G allocations: 1.929 TiB, 8.78% gc time, 87.66% compilation time: 0% of which was recompilation)
     Testing SciMLSensitivity tests passed
```

mohamed82008:
Looks good. Let's wait for CI to pass before we merge this PR though.

devmotion:
There's no response yet in DiffTests to my issue and PR, and I'm not sure when I will have time to dig more into other unrelated test errors on the master branch. So it might take a while until CI is fixed on the master branch.

mohamed82008:
Can we rebase this one now that DiffTests is upper bounded?

devmotion:
CI passed.

PR #209, opened 2022-10-01
Improve DiffRules integration and tests

devmotion:
This PR fixes some problems with the DiffRules integration and its tests. It is needed for https://github.com/JuliaDiff/DiffRules.jl/pull/79 (relevant DiffRules tests pass with that PR).

Mainly, the PR
- disables DiffRules definitions and tests for functions with complex outputs and derivatives as they are not supported by ReverseDiff
- adds support for `NaN` comparisons to the tests (necessary since in DiffRules undefined and non-existing derivatives are implemented as `NaN` and hence otherwise comparisons with ForwardDiff will fail if both return `NaN`)
- changes the tests such that they compare ReverseDiff results with the results of the corresponding ForwardDiff calls (currently, e.g., derivatives wrt to two arguments in ReverseDiff are compared with two separately computed ForwardDiff derivatives even though internally both derivatives are computed with one ForwardDiff call - this causes test errors since e.g. if one derivative is `NaN` the ForwardDiff results of both approaches are different for the derivative of the other argument, one will be `NaN` and one might not)
- improves the `map`/`broadcast`ing of DiffRules as currently internally derivatives are computed with ForwardDiff always for both arguments, even if only one is tracked (this was uncovered by the changes to the tests mentioned above and it ensures that derivatives of functions where derivatives are defined only for one argument return non-`NaN` results, as in ForwardDiff)

~~The `vcat` test error is unrelated and also present on the master branch and other PRs.~~ Edit: Fixed on the master branch,

I also assume we could do better than ForwardDiff here and also avoid that all results become `NaN` if derivatives are computed with respect to both arguments and only one is defined/exists.
But replacing ForwardDiff with a direct implementation of the DiffRules-derivatives seemed to require much larger changes, and I tried to apply only a somewhat minimal set of changes required for https://github.com/JuliaDiff/DiffRules.jl/pull/79.

### DIFF
--- a/src/derivatives/elementwise.jl
+++ b/src/derivatives/elementwise.jl
@@ -146,7 +146,6 @@ for (g!, g) in ((:map!, :map), (:broadcast!, :broadcast))
             end
             results = $(g)(df, value(x), value(y))
             map!(DiffResult.value, value(out), results)
-            out = track(DiffResults.value.(results), D, tp)
             cache = (results, df, index_bound(x, out), index_bound(y, out))
             record!(tape(y), SpecialInstruction, $(g), (x, y), out, cache)
             return out
--- a/src/derivatives/elementwise.jl
+++ b/src/derivatives/elementwise.jl
@@ -72,6 +72,7 @@ for g! in (:map!, :broadcast!), (M, f, arity) in DiffRules.diffrules(; filter_mo
         @warn "$M.$f is not available and hence rule for it can not be defined"
         continue  # Skip rules for methods not defined in the current scope
     end
+    (M, f) in SKIPPED_DIFFRULES && continue
     if arity == 1
         @eval @inline Base.$(g!)(f::typeof($M.$f), out::TrackedArray, t::TrackedArray) = $(g!)(ForwardOptimize(f), out, t)
     elseif arity == 2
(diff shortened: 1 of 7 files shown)


devmotion:
Bump :slightly_smiling_face: 

It would be good to fix ReverseDiff such that we can move forward with https://github.com/JuliaDiff/DiffRules.jl/pull/79.

mohamed82008:
Sorry for the delay, been swamped recently. I will take a look tonight.

mohamed82008 (commented):
Do we have tests for ForwardOptimize where both x and y are tracked? Seems there might be a method ambiguity error in this case?

mohamed82008:
Ah the methods are defined a bit further in the same file, nevermind.

src/derivatives/elementwise.jl
```diff
@@ -122,23 +123,54 @@ for (g!, g) in ((:map!, :map), (:broadcast!, :broadcast))
...
+        @eval function Base.$(g!)(f::ForwardOptimize{F}, out::TrackedArray, x::$A, y::TrackedReal{Y,D}) where {F,Y,D}
+            result = DiffResults.DiffResult(zero(Y), zero(D))
+            df = let result=result
+                (vx, vy) -> let vx=vx
+                    ForwardDiff.derivative!(result, s -> f.f(vx, s), vy)
+                end
+            end
             results = $(g)(df, value(x), value(y))
             map!(DiffResult.value, value(out), results)
+            out = track(DiffResults.value.(results), D, tp)
```
mohamed82008, line 149: this function is inplace so the input `out` should be over-written?

mohamed82008: I believe the over-writing is happening in the previous line so this line seems unnecessary and potentially buggy

src/derivatives/elementwise.jl
```diff
@@ -122,23 +123,54 @@ for (g!, g) in ((:map!, :map), (:broadcast!, :broadcast))
...
             cache = (results, df, index_bound(x, out), index_bound(y, out))
-            record!(tape(x, y), SpecialInstruction, $(g), (x, y), out, cache)
+            record!(tape(y), SpecialInstruction, $(g), (x, y), out, cache)
+            return out
+        end
+        @eval function Base.$(g!)(f::ForwardOptimize{F}, out::TrackedArray, x::TrackedArray{X}, y::$A) where {F,X}
+            result = DiffResults.GradientResult(SVector(zero(X)))
+            df = (vx, vy) -> let vy=vy
+                ForwardDiff.gradient!(result, s -> f.f(s[1], vy), SVector(vx))
```
mohamed82008, line 157: why `s[1]`?

src/derivatives/elementwise.jl
```diff
@@ -122,23 +123,54 @@ for (g!, g) in ((:map!, :map), (:broadcast!, :broadcast))
...
+            results = $(g)(df, value(x), value(y))
+            map!(DiffResult.value, value(out), results)
+            cache = (results, df, index_bound(x, out), index_bound(y, out))
+            record!(tape(x), SpecialInstruction, $(g), (x, y), out, cache)
+            return out
+        end
+        @eval function Base.$(g!)(f::ForwardOptimize{F}, out::TrackedArray, x::$A, y::TrackedArray{Y}) where {F,Y}
+            result = DiffResults.GradientResult(SVector(zero(Y)))
+            df = let vx=vx
+                (vx, vy) -> ForwardDiff.gradient!(result, s -> f.f(vx, s[1]), SVector(vy))
```
mohamed82008, line 168: `s[1]`?

src/derivatives/elementwise.jl
```diff
@@ -207,20 +240,52 @@ for g in (:map, :broadcast)
...
+            tp = tape(y)
+            out = track(DiffResults.value.(results), D, tp)
+            cache = (results, df, index_bound(x, out), index_bound(y, out))
+            record!(tp, SpecialInstruction, $(g), (x, y), out, cache)
+            return out
+        end
+        @eval function Base.$(g)(f::ForwardOptimize{F}, x::TrackedArray{X,D}, y::$A) where {F,X,D}
+            result = DiffResults.GradientResult(SVector(zero(X)))
+            df = (vx, vy) -> let vy=vy
+                ForwardDiff.gradient!(result, s -> f.f(s[1], vy), SVector(vx))
```
mohamed82008, line 275: `s[1]`?

src/derivatives/elementwise.jl
```diff
@@ -311,12 +383,25 @@ end
...
         else
-            istracked(a) && diffresult_increment_deriv!(a, output_deriv, results, 1, a_bound)
-            istracked(b) && diffresult_increment_deriv!(b, output_deriv, results, 2, b_bound)
+            if istracked(a)
+                p += 1
+                diffresult_increment_deriv!(a, output_deriv, results, p, a_bound)
+            end
+            if istracked(b)
+                p += 1
+                diffresult_increment_deriv!(b, output_deriv, results, p, b_bound)
```
mohamed82008, line 403: Same p comment

src/derivatives/elementwise.jl
```diff
@@ -207,20 +240,52 @@ for g in (:map, :broadcast)
...
+            tp = tape(x)
+            out = track(DiffResults.value.(results), D, tp)
+            cache = (results, df, index_bound(x, out), index_bound(y, out))
+            record!(tp, SpecialInstruction, $(g), (x, y), out, cache)
+            return out
+        end
+        @eval function Base.$(g)(f::ForwardOptimize{F}, x::$A, y::TrackedArray{Y,D}) where {F,Y,D}
+            result = DiffResults.GradientResult(SVector(zero(Y)))
+            df = (vx, vy) -> let vx=vx
+                ForwardDiff.gradient!(result, s -> f.f(vx, s[1]), SVector(vy))
```
mohamed82008, line 287: Same `s[1]` comment

src/derivatives/elementwise.jl
devmotion: Hmm good catch. I'm not sure, maybe this line ended up in the PR unintentionally. At least I don't remember why I added it. Let's see if anything breaks if I remove it.

devmotion:
```suggestion
```

src/derivatives/elementwise.jl
devmotion: Same as above.

src/derivatives/elementwise.jl
devmotion: Same as above.

src/derivatives/elementwise.jl
devmotion: Same as above.

src/derivatives/elementwise.jl
```diff
@@ -311,12 +383,25 @@ end
...
+            if istracked(b)
+                p += 1
+                diffresult_increment_deriv!(b, output_deriv, results, p)
+            end
         else
-            istracked(a) && diffresult_increment_deriv!(a, output_deriv, results, 1, a_bound)
-            istracked(b) && diffresult_increment_deriv!(b, output_deriv, results, 2, b_bound)
+            if istracked(a)
+                p += 1
+                diffresult_increment_deriv!(a, output_deriv, results, p, a_bound)
```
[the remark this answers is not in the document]
devmotion: To extract the correct partial. If `a` is tracked, its corresponding partial has index `p = 1` but if only `b` is tracked, the first partial (`p = 1`) corresponds to `b`. And if both `a` and `b` are tracked, `p = 1` corresponds to `a` and `p = 2` to `b`. So incrementing `p` in the branches allows us to avoid checking and handling all three scenarios separately.

Note that on the master branch `p = 1` for `a` and `p = 2` for `b` are hardcoded. That only works because on the master branch *always* the partials wrt to both arguments are computed and stored, even if only one argument is tracked.

src/derivatives/elementwise.jl
devmotion: Same as above.

mohamed82008 (approved):
LGTM

PR #210, opened 2022-10-02
Fix `vcat` method ambiguity issues on Julia 1.8

devmotion:
Fixes the `vcat` test errors due to a method ambiguity issue on Julia 1.8. These test errors are present on the master branch and show also up in other unrelated PRs.

Maybe at some point one should clean up the `vcat`/`hcat`/`cat` definitions, similar to https://github.com/FluxML/Tracker.jl/pull/128. But for now it seemed easier to just fix the reported method ambiguity issue.


devmotion:
The test errors are known and caused by the latest release of DiffTests. I opened an issue and a PR a week ago.

mohamed82008:
Can we upper bound the version of DiffTests?

devmotion:
A bit unfortunate but I guess that's probably the fastest and easiest solution for now. I also checked https://github.com/JuliaDiff/DiffTests.jl/pull/15 with this PR locally, and while it seems to fix the current issues it still fails due to some newly uncovered ambiguity issues (e.g., `*(::Diagonal, ::TrackedMatrix)`).

devmotion:
As expected, with DiffTests 0.1.1 all tests pass.

mohamed82008 (approved):

PR #227, opened 2023-04-28
Handle matrix times matrix = vector case

dkarrasch:
This "fixes" some weird behavior in the multiplication code. It occurred in https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/8a3027b_vs_960870e/PositiveFactorizations.primary.log PositiveFactorizations.jl. In their case, they want to multiply a matrix by the transpose of a row matrix (i.e., vector-like) into a vector. "By chance", this works out from the pov of dimensions, but from the pov of types having such a `mul!` method is weird and we may not wish to continue to support this (https://github.com/JuliaLang/julia/pull/49521#discussion_r1178114739). To make this package (and PositiveFactorizations.jl) run smoothly across (past and upcoming) versions, I proposed to simply catch that case and reshape the output vector to a matrix. In fact, this may even turn out to be advantageous in terms of performance, because that strange method in LinearAlgebra.jl calls `generic_matmatmul!`, for arguments matching the following signature:

```julia
mul!(::Vector{Float64}, ::Matrix{Float64}, ::Transpose{Float64, Matrix{Float64}}, ::Bool, ::Bool)
```

Once we reshape the vector to a matrix, this would go down the BLAS route!

### DIFF
--- a/src/derivatives/linalg/arithmetic.jl
+++ b/src/derivatives/linalg/arithmetic.jl
@@ -270,11 +270,13 @@ end
 # a * b
 
 function reverse_mul!(output, output_deriv, a, b, a_tmp, b_tmp)
-    istracked(a) && increment_deriv!(a, mul!(a_tmp, output_deriv, transpose(value(b))))
-    istracked(b) && increment_deriv!(b, mul!(b_tmp, transpose(value(a)), output_deriv))
-end
-function reverse_mul!(output, output_deriv::AbstractMatrix, a, b::AbstractMatrix, a_tmp::AbstractVector, b_tmp)
-    istracked(a) && increment_deriv!(a, mul!(reshape(a_tmp, :, 1), output_deriv, transpose(value(b))))
+    if istracked(a)
+        if a_tmp isa AbstractVector && b isa AbstractMatrix
+            increment_deriv!(a, mul!(reshape(a_tmp, :, 1), output_deriv, transpose(value(b))))
+        else
+            increment_deriv!(a, mul!(a_tmp, output_deriv, transpose(value(b))))
+        end
+    end
     istracked(b) && increment_deriv!(b, mul!(b_tmp, transpose(value(a)), output_deriv))
 end
 
--- a/src/derivatives/linalg/arithmetic.jl
+++ b/src/derivatives/linalg/arithmetic.jl
@@ -272,6 +272,10 @@ end
 function reverse_mul!(output, output_deriv, a, b, a_tmp, b_tmp)
     if istracked(a)
         if a_tmp isa AbstractVector && b isa AbstractMatrix
+            # this branch is required for scalar-valued functions that
+            # involve outer-products of vectors, for such functions, the target
+            # a_temp is a vector, but when b is a matrix, we cannot multiply into a vector,
+            # so need to reshape memory to look like matrix (see PositiveFactorizations.jl)
             increment_deriv!(a, mul!(reshape(a_tmp, :, 1), output_deriv, transpose(value(b))))
         else
             increment_deriv!(a, mul!(a_tmp, output_deriv, transpose(value(b))))
@@ -285,8 +289,14 @@ for (f, F) in ((:transpose, :Transpose), (:adjoint, :Adjoint))
         # a * f(b)
         function reverse_mul!(output, output_deriv, a, b::$F, a_tmp, b_tmp)
             _b = ($f)(b)
-            istracked(a) && increment_deriv!(a, mul!(a_tmp, output_deriv, mulargvalue(b)))
-            istracked(_b) && increment_deriv!(_b, ($f)(mul!(b_tmp, ($f)(output_deriv), value(a))))
+            if istracked(a)
+                if a_tmp isa AbstractVector
+                    increment_deriv!(a, mul!(reshape(a_tmp, :, 1), output_deriv, mulargvalue(_b)))
+                else
+                    increment_deriv!(a, mul!(a_tmp, output_deriv, mulargvalue(b)))
+                end
+            end
+            istracked(_b) && increment_deriv!(_b, ($f)(mul!(($f)(b_tmp), ($f)(output_deriv), value(a))))
         end
            # f(a) * b
         function reverse_mul!(output, output_deriv, a::$F, b, a_tmp, b_tmp)
--- a/src/derivatives/linalg/arithmetic.jl
+++ b/src/derivatives/linalg/arithmetic.jl
@@ -273,6 +273,10 @@ function reverse_mul!(output, output_deriv, a, b, a_tmp, b_tmp)
     istracked(a) && increment_deriv!(a, mul!(a_tmp, output_deriv, transpose(value(b))))
     istracked(b) && increment_deriv!(b, mul!(b_tmp, transpose(value(a)), output_deriv))
 end
+function reverse_mul!(output, output_deriv::AbstractMatrix, a, b::AbstractMatrix, a_tmp::AbstractVector, b_tmp)
+    istracked(a) && increment_deriv!(a, mul!(reshape(a_tmp, :, 1), output_deriv, transpose(value(b))))
+    istracked(b) && increment_deriv!(b, mul!(b_tmp, transpose(value(a)), output_deriv))
+end
 
 for (f, F) in ((:transpose, :Transpose), (:adjoint, :Adjoint))
     @eval begin
--- a/test/api/GradientTests.jl
+++ b/test/api/GradientTests.jl
@@ -188,12 +188,22 @@ for f in DiffTests.VECTOR_TO_NUMBER_FUNCS
 end
 
 # PR #227
-function norm_hermitian(v)
-    A = I - 2 * v * v'
-    return norm(A' * A)
-end
-test_println("VECTOR_TO_NUMBER_FUNCS", norm_hermitian)
-test_unary_gradient(norm_hermitian, rand(5))
+norm_hermitian1(v) = (A = I - 2 * v * v'; norm(A' * A))
+norm_hermitian2(v) = (A = I - 2 * v * transpose(v); norm(transpose(A) * A))
+norm_hermitian3(v) = (A = I - 2 * v * collect(v'); norm(collect(A') * A))
+norm_hermitian4(v) = (A = I - 2 * v * v'; norm(transpose(A) * A))
+norm_hermitian5(v) = (A = I - 2 * v * transpose(v); norm(A' * A))
+
+test_println("VECTOR_TO_NUMBER_FUNCS", norm_hermitian1)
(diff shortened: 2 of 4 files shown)


dkarrasch:
Gentle bump.

devmotion:
I didn't get why exactly this method is needed. Shouldn't the re-routing happen in LinearAlgebra, why is some special handling in ReverseDiff needed?

In any case, it seems the code is not covered by tests yet.

dkarrasch:
So here's the full stacktrace. It shows that the output is allocated by ReverseDiff, and then LinearAlgebra takes what it gets:

```julia
MethodError: no method matching mul!(::Vector{Float64}, ::Matrix{Float64}, ::Transpose{Float64, Matrix{Float64}}, ::Bool, ::Bool)
  
  Closest candidates are:
    mul!(::StridedVector{T}, ::StridedVecOrMat{T}, !Matched::StridedVector{T}, ::Number, ::Number) where T<:Union{Float32, Float64, ComplexF64, ComplexF32}
     @ LinearAlgebra /opt/julia/share/julia/stdlib/v1.10/LinearAlgebra/src/matmul.jl:66
    mul!(!Matched::StridedMatrix{T}, ::StridedVecOrMat{T}, ::Transpose{<:Any, <:StridedVecOrMat{T}}, ::Number, ::Number) where T<:Union{Float32, Float64, ComplexF64, ComplexF32}
     @ LinearAlgebra /opt/julia/share/julia/stdlib/v1.10/LinearAlgebra/src/matmul.jl:369
    mul!(!Matched::StridedMatrix{T}, ::Union{Adjoint{<:Any, <:StridedVecOrMat{T}}, Transpose{<:Any, <:StridedVecOrMat{T}}, StridedMatrix{T}, StridedVector{T}}, ::Union{Adjoint{<:Any, <:StridedVecOrMat{T}}, Transpose{<:Any, <:StridedVecOrMat{T}}, StridedMatrix{T}, StridedVector{T}}, ::Number, ::Number) where T<:Union{Float32, Float64, ComplexF64, ComplexF32}
     @ LinearAlgebra /opt/julia/share/julia/stdlib/v1.10/LinearAlgebra/src/matmul.jl:356
    ...
  
  Stacktrace:
    [1] mul!
      @ /opt/julia/share/julia/stdlib/v1.10/LinearAlgebra/src/matmul.jl:251 [inlined]
    [2] reverse_mul!(output::ReverseDiff.TrackedArray{Float64, Float64, 2, Matrix{Float64}, Matrix{Float64}}, output_deriv::Matrix{Float64}, a::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}, b::Matrix{ReverseDiff.TrackedReal{Float64, Float64, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}}}, a_tmp::Vector{Float64}, b_tmp::Matrix{Float64})
      @ ReverseDiff ~/.julia/packages/ReverseDiff/Zu4v6/src/derivatives/linalg/arithmetic.jl:273
    [3] special_reverse_exec!(instruction::ReverseDiff.SpecialInstruction{typeof(*), Tuple{ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}, Matrix{ReverseDiff.TrackedReal{Float64, Float64, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}}}}, ReverseDiff.TrackedArray{Float64, Float64, 2, Matrix{Float64}, Matrix{Float64}}, Tuple{Vector{Float64}, Matrix{Float64}}})
      @ ReverseDiff ~/.julia/packages/ReverseDiff/Zu4v6/src/derivatives/linalg/arithmetic.jl:265
    [4] reverse_exec!(instruction::ReverseDiff.SpecialInstruction{typeof(*), Tuple{ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}, Matrix{ReverseDiff.TrackedReal{Float64, Float64, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}}}}, ReverseDiff.TrackedArray{Float64, Float64, 2, Matrix{Float64}, Matrix{Float64}}, Tuple{Vector{Float64}, Matrix{Float64}}})
      @ ReverseDiff ~/.julia/packages/ReverseDiff/Zu4v6/src/tape.jl:93
    [5] reverse_pass!(tape::Vector{ReverseDiff.AbstractInstruction})
      @ ReverseDiff ~/.julia/packages/ReverseDiff/Zu4v6/src/tape.jl:87
    [6] reverse_pass!
      @ ~/.julia/packages/ReverseDiff/Zu4v6/src/api/tape.jl:36 [inlined]
    [7] seeded_reverse_pass!(result::Vector{Float64}, output::ReverseDiff.TrackedReal{Float64, Float64, Nothing}, input::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}, tape::ReverseDiff.GradientTape{var"#8#15"{DataType, var"#5#12"}, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}, ReverseDiff.TrackedReal{Float64, Float64, Nothing}})
      @ ReverseDiff ~/.julia/packages/ReverseDiff/Zu4v6/src/api/utils.jl:31
    [8] seeded_reverse_pass!(result::Vector{Float64}, t::ReverseDiff.GradientTape{var"#8#15"{DataType, var"#5#12"}, ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}, ReverseDiff.TrackedReal{Float64, Float64, Nothing}})
      @ ReverseDiff ~/.julia/packages/ReverseDiff/Zu4v6/src/api/tape.jl:47
    [9] gradient(f::Function, input::Vector{Float64}, cfg::ReverseDiff.GradientConfig{ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}})
      @ ReverseDiff ~/.julia/packages/ReverseDiff/Zu4v6/src/api/gradients.jl:24
   [10] gradient(f::Function, input::Vector{Float64})
      @ ReverseDiff ~/.julia/packages/ReverseDiff/Zu4v6/src/api/gradients.jl:22
```

I'm not familiar with AD and how to set it up so that it goes this route. As I wrote, it came up with PositiveFactorizations.jl and AD-ing `cholesky` or something like that, so not easily reproducible here.

> In any case, it seems the code is not covered by tests yet.

True, there's quite a lot uncovered in that code area. Any help with that would be much appreciated.

test/derivatives/LinAlgTests.jl
```diff
@@ -223,8 +223,11 @@ for f in (
     test_arr2num(f, x, tp)
 end
 
+vec_to_hermitian = (v) -> begin A = I - 2 * v * collect(v'); A = collect(A') * A end;
```
devmotion, line 226:
```suggestion
# PR #227
function norm_hermitian(v)
    A = I - 2 * v * v'
    return norm(A' * A)
end
```

test/derivatives/LinAlgTests.jl
```diff
@@ -223,8 +223,11 @@ for f in (
...
+
 for f in (
     y -> vec(y)' * Matrix{Float64}(I, length(y), length(y)) * vec(y),
+    y -> norm(vec_to_hermitian(y)),
```
devmotion, line 230:
```suggestion
    norm_hermitian,
```

devmotion (commented):
I checked and it seems that ReverseDiff computes gradients for these functions even on the master branch? Are you sure the test is correct?

dkarrasch:
I think it no longer does after I merged https://github.com/JuliaLang/julia/pull/49521.

devmotion:
It seems that the PR introduces method ambiguity issues which cause test failures after removing `collect` in the tests.

Can you fix these and add also a version of `norm_hermitian` with `collect` (but keep the one without as well since it seems to cover some of the method ambiguity issues)?

If I can find some time I'll also check if the initial problem can be fixed in some other way, since evidently adding new dispatches seems a bit problematic.

dkarrasch:
Interestingly, the ambiguities do not occur on v1.8, only on the very old versions. I have now avoided adding another method and introduced branches in the existing one. Those branches, however, should be compiled away because the conditions can be checked in the type domain.

BTW, this packages has more than 3k ambiguities! My computer almost crashed when running Aqua.jl on it.

src/derivatives/linalg/arithmetic.jl
```diff
@@ -270,7 +270,13 @@ end
 # a * b
 
 function reverse_mul!(output, output_deriv, a, b, a_tmp, b_tmp)
-    istracked(a) && increment_deriv!(a, mul!(a_tmp, output_deriv, transpose(value(b))))
+    if istracked(a)
+        if a_tmp isa AbstractVector && b isa AbstractMatrix
+            increment_deriv!(a, mul!(reshape(a_tmp, :, 1), output_deriv, transpose(value(b))))
```
devmotion, line 275: It seems this branch still causes test failures on some Julia versions - does it call a `mul!` method that was added only recently?

test/api/GradientTests.jl
```diff
@@ -187,6 +187,24 @@ for f in DiffTests.VECTOR_TO_NUMBER_FUNCS
...
+test_println("VECTOR_TO_NUMBER_FUNCS", norm_hermitian1)
+test_unary_gradient(norm_hermitian1, rand(5))
+test_println("VECTOR_TO_NUMBER_FUNCS", norm_hermitian2)
+test_unary_gradient(norm_hermitian2, rand(5))
+test_println("VECTOR_TO_NUMBER_FUNCS", norm_hermitian3)
+test_unary_gradient(norm_hermitian3, rand(5))
+test_println("VECTOR_TO_NUMBER_FUNCS", norm_hermitian4)
+test_unary_gradient(norm_hermitian4, rand(5))
+test_println("VECTOR_TO_NUMBER_FUNCS", norm_hermitian5)
+test_unary_gradient(norm_hermitian5, rand(5))
```
devmotion, line 206: Slightly more compact:
```suggestion
for f in (norm_hermitian1, norm_hermitian2, norm_hermitian3, norm_hermitian4, norm_hermitian5)
    test_println("VECTOR_TO_NUMBER_FUNCS", f)
    test_unary_gradient(f, rand(5))
end
```

src/derivatives/linalg/arithmetic.jl
dkarrasch: It's not this line, it's line 288 below, where the same error occurs and requires the same fix.

devmotion: Ah sorry, I misread the error message then.

devmotion (approved):

devmotion:
Thank you @dkarrasch!

Issue #226, opened 2023-04-25
ReverseDiff defines a huge number of methods.


KristofferC:
Loading ReverseDiff and recording what methods are inserted we can see that this package defines a very large number of methods:

<img width="639" alt="image" src="https://user-images.githubusercontent.com/1282691/234338028-ec3f0064-4a08-4e1a-920d-3cc76c11a83b.png">

For example, there are almost a thousand definitions for `materialize` (which is not even something that should be extended according to the interface manual: https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting).

There are also many many for `hcat`, `vcat`.

Structuring the package like this heavily pessimize its load time and it would be good to be less "brute force" than this way where you just write a heavily nested loop and define methods in it.

prbzrg:
I think generalizing the methods definitions is the actual problem here.
```julia
julia> (xs::typeof(vcat))(x::String, s::String) = println("working")

julia> vcat("hi", "bye")
working

julia> (xs::Union{typeof(vcat), typeof(hcat)})(x::String, s::String) = println("working")
ERROR: Method dispatch is unimplemented currently for this method signature
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1

```
If we could redirect a group of methods by arguments types, we wouldn't need to have a macro loop.

https://github.com/JuliaDiff/ReverseDiff.jl/blob/master/src/derivatives/arrays.jl#L51-L54

prbzrg:
e.g. It would be:
```julia
(fun::AllArrayFunctionsType)(xs::Vararg{Any, TrackedVecOrMat}) = track(fun, xs...)
```

PR #231, opened 2023-05-16
Add `Diagonal` in Array types

### LINKED ISSUE #223: `MethodError: *(::Diagonal, ::ReverseDiff.TrackedArray)` is ambiguous.
I got the following error when using `ReverseDiff`
```
using LinearAlgebra, ReverseDiff
n = 30
D = Diagonal([0.1 + 0.9 * (i - 1) / (n - 1) for i = 1:n])
F = x -> sqrt.(D) * (x .- 1)
x = ones(n)
ReverseDiff.gradient(x -> dot(F(x), x), x)
```
and got the following error
```
ERROR: MethodError: *(::Diagonal{Float64, Vector{Float64}}, ::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}) is ambiguous. Candidates:
  *(x::AbstractMatrix, y::ReverseDiff.TrackedArray{V, D, 1}) where {V, D} in ReverseDiff at .julia\packages\ReverseDiff\YkVxM\src\derivatives\linalg\arithmetic.jl:214
  *(x::AbstractArray, y::ReverseDiff.TrackedArray{V, D, 1}) where {V, D} in ReverseDiff at .julia\packages\ReverseDiff\YkVxM\src\derivatives\linalg\arithmetic.jl:214
  *(x::AbstractMatrix, y::ReverseDiff.TrackedArray{V, D}) where {V, D} in ReverseDiff at .julia\packages\ReverseDiff\YkVxM\src\derivatives\linalg\arithmetic.jl:214
  *(x::AbstractArray, y::ReverseDiff.TrackedArray{V, D}) where {V, D} in ReverseDiff at .julia\packages\ReverseDiff\YkVxM\src\derivatives\linalg\arithmetic.jl:214
  *(D::Diagonal, V::AbstractVector) in LinearAlgebra at julia-1.8.0\share\julia\stdlib\v1.8\LinearAlgebra\src\diagonal.jl:242
Possible fix, define
  *(::Diagonal, ::ReverseDiff.TrackedArray{V, D, 1}) where {V, D}
Stacktrace:
 [1] (::var"#3#4")(x::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}})
   @ Main .\REPL[6]:1
 [2] (::var"#7#8")(x::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}})
   @ Main .\REPL[9]:1
 [3] ReverseDiff.GradientTape(f::var"#7#8", input::Vector{Float64}, cfg::ReverseDiff.GradientConfig{ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}})
   @ ReverseDiff .julia\packages\ReverseDiff\YkVxM\src\api\tape.jl:199
 [4] gradient(f::Function, input::Vector{Float64}, cfg::ReverseDiff.GradientConfig{ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{Float64}, Vector{Float64}}}) (repeats 2 times)
   @ ReverseDiff .julia\packages\ReverseDiff\YkVxM\src\api\gradients.jl:22
 [5] top-level scope
   @ REPL[9]:1
```

tmigot:
Close #223

### DIFF
--- a/src/ReverseDiff.jl
+++ b/src/ReverseDiff.jl
@@ -23,7 +23,7 @@ using ChainRulesCore
 
 # Not all operations will be valid over all of these types, but that's okay; such cases
 # will simply error when they hit the original operation in the overloaded definition.
-const ARRAY_TYPES = (:AbstractArray, :AbstractVector, :AbstractMatrix, :Array, :Vector, :Matrix)
+const ARRAY_TYPES = (:AbstractArray, :AbstractVector, :AbstractMatrix, :Array, :Vector, :Matrix, :Diagonal)
 const REAL_TYPES = (:Bool, :Integer, :(Irrational{:ℯ}), :(Irrational{:π}), :Rational, :BigFloat, :BigInt, :AbstractFloat, :Real, :Dual)
 
 const SKIPPED_UNARY_SCALAR_FUNCS  = Symbol[:isinf, :isnan, :isfinite, :iseven, :isodd, :isreal, :isinteger]


devmotion:
I think, even though this PR might fix the issue (untested it seems) it's not the fix we want: The package already defines (too) many methods (see eg. https://github.com/JuliaDiff/ReverseDiff.jl/issues/226) and adding `Diagonal` to the array type union will increase this nummber further by adding many more definitions - even though the issue was only about a specific method for `*`. It also seems a bit surprising to treat `Diagonal` in such a special way but not e.g. `Adjoint`, `Symmetric` and other matrix types in LinearAlgebra.

tiemvanderdeure:
Has there been any progress on this? 

Multiplying a Cholesky matrix with a Diagonal does not work with ReverseDiff, e.g. in the example shared here: https://github.com/TuringLang/Turing.jl/issues/1870#issuecomment-1706595909

devmotion:
There's no multiplication of a `Diagonal` with a `Cholesky` matrix in the linked example? The model only multiplies a `Diagonal` with a `LowerTriangular` matrix. In any case, you could rewrite the multiplication with a `Diagonal` using broadcasting.

Generally, I still have the same feeling as in https://github.com/JuliaDiff/ReverseDiff.jl/pull/231#issuecomment-1550198020.

PR #232, opened 2023-05-20
Allow custom struct args to grad_from_chainrules macro

mohamed82008:
I noticed that we don't esc the types of the arguments in the `@grad_from_chainrules` macro. This PR fixes that.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -317,9 +317,18 @@ macro grad_from_chainrules(fcall)
     Meta.isexpr(fcall, :call) && length(fcall.args) >= 2 ||
         error("`@grad_from_chainrules` has to be applied to a function signature")
     f = esc(fcall.args[1])
-    xs = fcall.args[2:end]
+    xs = map(fcall.args[2:end]) do x
+        if x isa Expr && x.head == :(::)
+            if length(x.args) == 1 # ::T without var name
+                return :($(gensym())::$(esc(x.args[1])))
+            else # x::T
+                return :($(x.args[1])::$(esc(x.args[2])))
+            end
+        else
+            return x
+        end
+    end
     args_l, args_r, args_track, args_fixed, arg_types, kwargs = _make_fwd_args(f, xs)
-
     return quote
         $f($(args_l...)) = ReverseDiff.track($(args_r...))
         function ReverseDiff.track($(args_track...))
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -318,7 +318,7 @@ macro grad_from_chainrules(fcall)
         error("`@grad_from_chainrules` has to be applied to a function signature")
     f = esc(fcall.args[1])
     xs = map(fcall.args[2:end]) do x
-        if x isa Expr && x.head == :(::)
+        if Meta.isexpr(x, :(::))
             if length(x.args) == 1 # ::T without var name
                 return :($(gensym())::$(esc(x.args[1])))
             else # x::T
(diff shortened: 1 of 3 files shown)


test/ChainRulesTests.jl
```diff
@@ -6,6 +6,11 @@ using DiffResults
 using ReverseDiff
 using Test
 
+const NoTangent = ChainRulesCore.NoTangent
+
```
devmotion, line 10: This should be unnecessary due to the `using ChainRulesCore` above:
```suggestion
```

src/macros.jl
```diff
@@ -317,9 +317,18 @@ macro grad_from_chainrules(fcall)
     Meta.isexpr(fcall, :call) && length(fcall.args) >= 2 ||
         error("`@grad_from_chainrules` has to be applied to a function signature")
     f = esc(fcall.args[1])
-    xs = fcall.args[2:end]
+    xs = map(fcall.args[2:end]) do x
+        if x isa Expr && x.head == :(::)
```
devmotion, line 321:
```suggestion
        if Meta.isexpr(x, :(::))
```

src/macros.jl
```diff
@@ -317,9 +317,18 @@ macro grad_from_chainrules(fcall)
...
+            if length(x.args) == 1 # ::T without var name
+                return :($(gensym())::$(esc(x.args[1])))
```
devmotion, line 323: Why is a variable name needed? Can't we just escape the type?

src/macros.jl
```diff
@@ -317,9 +317,18 @@ macro grad_from_chainrules(fcall)
...
+            else # x::T
```
devmotion, line 324: Would be safer to check that `length(x.args) == 2` and error otherwise.

mohamed82008: Do you know of a case where head is `:(::)` and args are more than 2? It's a syntax error in Julia now if I try to construct the expression manually and eval it.

mohamed82008: if it's a syntax error, I can't test the third else branch

src/macros.jl
mohamed82008: it was sure removing the var nam errored but it now works, let's see if works on CI

mohamed82008: I see. Removing the gensym breaks this line
```julia
$f($(args_l...)) = ReverseDiff.track($(args_r...))
```
making it throw the following syntax error

ERROR: syntax: invalid "::" syntax around ~/.julia/dev/ReverseDiff/src/macros.jl:338
Stacktrace:
 [1] top-level scope

mohamed82008: I am not sure why this errors, seems like valid Julia syntax to me.

devmotion: Seems to me the problem is not the LHS but rather the RHS: For instance, with the gensymed variable removed, we get
```julia
julia> @macroexpand ReverseDiff.@grad_from_chainrules sin(x::Float64)
quote
    #= /home/david/.julia/dev/ReverseDiff/src/macros.jl:333 =#
    sin(var"#50#x"::Float64) = begin
            #= /home/david/.julia/dev/ReverseDiff/src/macros.jl:333 =#
            (ReverseDiff.ReverseDiff).track(sin, var"#50#x"::Float64)
        end
...

julia> @macroexpand ReverseDiff.@grad_from_chainrules sin(::Float64)
quote
    #= /home/david/.julia/dev/ReverseDiff/src/macros.jl:333 =#
    sin(::Float64) = begin
            #= /home/david/.julia/dev/ReverseDiff/src/macros.jl:333 =#
            (ReverseDiff.ReverseDiff).track(sin, ::Float64)
        end
...
```
Seems like a bug in `_make_fwd_args` that we include the types in the `track` calls but in any case it seems we need the gensymed variables.

mohamed82008:
Thanks @devmotion. I think this is probably good to go.

devmotion (approved):

Issue #242, opened 2023-10-27
  MethodError: vcat(::ReverseDiff.TrackedArray{Float32, Float32, 2, Matrix{Float32}, Matrix{Float32}}, ::Matrix{Float32}) is ambiguous.


prbzrg:
I get this error on Julia 1.10, it's working on Julia 1.9 or if I use `cat(...; dims=1)` instead.

```
  MethodError: vcat(::ReverseDiff.TrackedArray{Float32, Float32, 2, Matrix{Float32}, Matrix{Float32}}, ::Matrix{Float32}) is ambiguous.
  
  Candidates:
    vcat(x::ReverseDiff.TrackedArray{V, D, 2} where {V, D}, xs::AbstractMatrix...)
      @ ReverseDiff ~/.julia/packages/ReverseDiff/UJhiD/src/derivatives/arrays.jl:53
    vcat(X::Union{Number, AbstractVecOrMat{<:Number}}...)
      @ SparseArrays /opt/hostedtoolcache/julia/1.10.0-beta3/x64/share/julia/stdlib/v1.10/SparseArrays/src/sparsevector.jl:1229
    vcat(x::ReverseDiff.TrackedArray{V, D, 2} where {V, D}, xs::AbstractVecOrMat...)
      @ ReverseDiff ~/.julia/packages/ReverseDiff/UJhiD/src/derivatives/arrays.jl:53
  
  Possible fix, define
    vcat(::ReverseDiff.TrackedArray{V, D, 2} where {V, D}, ::Vararg{AbstractMatrix{<:Number}})
  
```

https://github.com/impICNF/ContinuousNormalizingFlows.jl/actions/runs/6660418729/job/18101501796#step:5:18373

devmotion:
```julia
    vcat(X::Union{Number, AbstractVecOrMat{<:Number}}...)
```
This method signature seems pretty bad, it's incredibly generic and even covers `vcat()`. Maybe this is actually a SparseArrays issue?

prbzrg:
Yes, I agree. But I think arrays should have an eltype in the definitions. e.g. `AbstractMatrix{<:Real}`.

https://github.com/JuliaDiff/ReverseDiff.jl/blob/b669294586b0789ab10bbda223543bcec64cf8aa/src/derivatives/arrays.jl#L43-L54

prbzrg:
CI failed too. so I close the PR.
[...]

PR #244, opened 2023-10-30
Update CI julia versions

prbzrg:
This reveals the problem in #242

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,13 +23,10 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - '1.0'
           - '1.6'
-          # - '1.7'
-          # - '1.8'
-          # - '1.9'
           - '1'
           - '~1.10.0-0'
-          # - 'nightly'
         os:
           - ubuntu-latest
         arch:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -23,9 +23,13 @@ jobs:
...
         version:
-          - '1.0'
-          - '1.3'
+          - '1.6'
+          # - '1.7'
+          # - '1.8'
+          # - '1.9'
           - '1'
+          - '~1.10.0-0'
+          # - 'nightly'
```
devmotion, line 32: IMO we should test the stable Julia version (1.9 currently, `1` in CI), the latest LTS (`1.6`), and the oldest supported version (`'1.0'`) to avoid accidentally breaking anything.

We could add nightly but I assume it would be quite noisy - nightly tends to break things quite often (in particular when relying on internals such as LinearAlgebra here) but most of these broken nightlies won't result in a broken stable release in my experience.

devmotion:
Is the main motivation for this PR the issue in #242? I guess we don't need to run CI to reproduce it, maybe rather add a MWE in #242?

prbzrg:
this line:
https://github.com/JuliaDiff/ReverseDiff.jl/blob/b669294586b0789ab10bbda223543bcec64cf8aa/test/derivatives/ArrayFunctionTests.jl#L93

PR #249, opened 2024-02-13
(Temporarily) fix stable docs link in README.md

jacobusmmsmit:
#246 pointed out that the docs are not correctly deployed. Unfortunately, this is still the case so in the meantime a quick "bandaid" fix which makes the stable docs accessible is just to change the stable docs link.

Right now the stable docs are hard to find unless you know that removing `/stable` from the url goes there.

Apologies if this is the wrong fix to the problem, it was just bugging me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # ReverseDiff
 
-[![Stable](https://img.shields.io/badge/docs-stable-blue.svg)](https://juliadiff.org/ReverseDiff.jl/stable)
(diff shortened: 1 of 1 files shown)


gdalle (approved):
yeah it's been bugging me too

devmotion:
This PR has to be reverted. The new link points to completely outdated docs. IMO a broken link is better than one to incorrect docs.

gdalle:
Okay, but is there a link to the current stable docs? I thought this was the right one too

devmotion:
It's some docs from years ago.

PR #252, opened 2024-04-14
Fix tape printing

gdalle:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/1354?notification_referrer_id=NT_kwDOAVvVTrM5OTI3ODA3ODE2OjIyNzk1NTk4#issuecomment-2053902656

### DIFF
--- a/src/tape.jl
+++ b/src/tape.jl
@@ -110,7 +110,7 @@ function Base.show(io::IO, instruction::AbstractInstruction, pad = "")
 end
 
 function Base.show(io::IO, tp::InstructionTape)
-    println("$(length(tp))-element InstructionTape:")
+    println(io, "$(length(tp))-element InstructionTape:")
     i = 1
     for instruction in tp
         print(io, "$i => ")


src/tape.jl
```diff
@@ -110,7 +110,7 @@ function Base.show(io::IO, instruction::AbstractInstruction, pad = "")
 end
 
 function Base.show(io::IO, tp::InstructionTape)
-    println("$(length(tp))-element InstructionTape:")
+    println(io, "$(length(tp))-element InstructionTape:")
```
devmotion, line 113:
```suggestion
    println(io, length(tp), "-element InstructionTape:")
```

devmotion (commented):
Can you add a test (eg using `string` or an `IOBuffer`?)?

gdalle:
> Can you add a test (eg using string or an IOBuffer?)?

done

devmotion (approved):
Looks good to me, thanks!

PR #253, opened 2024-04-14
Revert "(Temporarily) fix stable docs link in README.md"

devmotion:
Reverts JuliaDiff/ReverseDiff.jl#249

Ref https://github.com/JuliaDiff/ReverseDiff.jl/pull/249#issuecomment-2053944705
Ref https://github.com/JuliaDiff/ReverseDiff.jl/pull/246

PR #255, opened 2024-04-16
Update TagBot config

devmotion:
@gdalle #253 didn't fix the stable docs build. https://github.com/JuliaDiff/ReverseDiff.jl/actions shows that the TagBot action executed successfully but did not trigger a docs build. I assume the remaining issue might be an outdated TagBot config. In this PR I copied the recommended TagBot config.

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,11 +1,31 @@
 name: TagBot
 on:
-  schedule:
-    - cron: 0 * * * *
+  issue_comment:
+    types:
+      - created
+  workflow_dispatch:
+    inputs:
+      lookback:
+        default: "3"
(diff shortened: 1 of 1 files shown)


devmotion:
I think this should be uncontroversial and only affects currently broken GH actions, so I'll merge and check whether it improves things.

gdalle:
I wanted to check the settings of the repo but I don't have access: are the documenter key and secret correctly set up? Otherwise TagBot won't succeed in triggering docs build.
In the meantime we can push a manual tag to build the docs

gdalle:
@devmotion should we push a Documenter tag to see if your setup worked?
https://documenter.juliadocs.org/stable/man/hosting/#Fixing-broken-release-deployments

PR #256, opened 2024-04-22
Format code with `JuliaFormatter`

mofeing:
This PR formats the code with `JuliaFormatter` and adds a couple of format checking workflows.


.github/workflows/format-check.yml
```diff
@@ -0,0 +1,41 @@
+name: format-check
+
+on:
+  push:
+    branches:
+      - 'master'
+    tags: '*'
+  pull_request:
```
devmotion, line 8: IMO the only thing that is needed is a check that is run in every PR. Maybe just copy common setups such as in https://github.com/JuliaDiff/ChainRulesCore.jl/blob/main/.github/workflows/format.yml?

.github/workflows/format-pr.yml
```diff
@@ -0,0 +1,51 @@
+name: format-pr
```
devmotion, line 1: I would not add such an action, and in particular I don't think it is useful to run it as a cron job. If the check fails in a PR, one can just push a commit with formatting fixes; and if only PRs with correct format are merged, there's no need to check or create PRs on the master branch.

devmotion (commented):
Before we format the code base, we have to agree on a code style. IMO we shouldn't rely on general defaults of JuliaFormatter but fix the style that's agreed on in a `.JuliaFormatter.toml` file in the repo.

I tend to use blue style (as e.g. ChainRulesCore: https://github.com/JuliaDiff/ChainRulesCore.jl/blob/main/.JuliaFormatter.toml) but there are of course other common styles.

.github/workflows/format-check.yml
mofeing: Oh nice. I didn't know about `reviewdog/action-suggester`. Just refactored to be like the `ChainRulesCore` workflow.

.github/workflows/format.yml
```diff
@@ -0,0 +1,26 @@
+name: format-check
+
+on:
+  push:
+    branches:
+      - 'master'
+    tags: '*'
```
devmotion, line 7: I think we could skip this and only run the check on PRs:
```suggestion
```

devmotion (commented):
Maybe add the blue style badge to the README? Otherwise it looks good to me 🙂

.github/workflows/format.yml
mofeing: Said and done!

PR #261, opened 2024-07-28
Realias DiffResults whenever possible

### LINKED ISSUE #251: DiffResults objects are not re-aliased properly
The DiffResults documentation clearly states that whenever a `DiffResult` object is used, it must be realiased at the end of the function call:

```julia
result = do_stuff!(result, args...)
```

See for instance the docstrings in https://juliadiff.org/DiffResults.jl/stable/#Mutating-a-DiffResult, or the issue [DiffResults#17](https://github.com/JuliaDiff/DiffResults.jl/issues/17)

This is not done by ReverseDiff, for instance here:

https://github.com/JuliaDiff/ReverseDiff.jl/blob/c982cde5494fc166965a9d04691f390d9e3073fd/src/api/gradients.jl#L21-L27

Am I right in deducing that it can lead to incorrectness?

gdalle:
Fix #251 partially.

- [x] Do `result = operator!(result, f, x)` instead of `operator!(result, f, x)` whenever possible (see https://github.com/JuliaDiff/DiffResults.jl/issues/27)
- [ ] There are still some parts involving closures which I don't know how to handle.
- [x] Adjust the tests accordingly
- [ ] Add new tests that would fail on the current release.

@devmotion do you know the package well enough to help me figure out what I missed?


devmotion:
I'm off until next week and probably won't be able to add a meaningful comment until then.

gdalle:
No worries, I'll wait for your opinion!

devmotion (commented):
At first glance, my impression is that there's more work to be done for full StaticArrays support than just fixing the incorrect use of DiffResults (see e.g. https://github.com/JuliaDiff/ReverseDiff.jl/pull/154).

I'm not a fan of major changes to tests in PRs with such significant changes either, so I suggest splitting the PR into multiple smaller PRs, e.g., one that adds and/or modifies tests (without changing the package) and one that fixes the DiffResults API usage.

Given that static arrays have additional problems, I would also suggest just testing `ImmutableDiffResult` and `MutableDiffResult` with `Array`s (and not `SArray`s or `MArray`s) initially (I assume that this will also cause problems currently if the DiffResults API is used incorrectly).

To increase readability, I think we might also want to use the bangbang notation (`!!`) for internal functions that are changed to be only sometimes mutable (I'd refrain from changing the user-facing API right now).

devmotion:
Fixed by #295

Issue #259, opened 2024-05-06
ERROR: LoadError: Some tests did not pass: 146 passed, 0 failed, 1 errored, 0 broken.


iajzenszmi:
![Screenshot from 2024-05-06 15-29-33](https://github.com/JuliaDiff/ReverseDiff.jl/assets/5051767/c0159a24-6535-4783-8b39-04e3eb9f9688)
ERROR: LoadError: Some tests did not pass: 146 passed, 0 failed, 1 errored, 0 broken.

devmotion:
I'll close this as a duplicate of #217 (which is caused by https://github.com/JuliaSparse/SparseArrays.jl/issues/431).

PR #262, opened 2024-07-29
Make CI pass

gdalle:
- Put all tests inside a big `@testset`
- Identify and mark failing `cat` tests as broken

At the moment, `testcat` in `ArrayFunctionTests.jl` is broken for 4 or more arguments because of a method ambiguity (see #242). This PR doesn't fix the source of the problem, it just ignores broken tests so that CI passes. To be fair, these tests are not very useful.

These broken tests pass on Julia 1.0 and 1.3 (for which support should be dropped anyway) but fail on 1.10. At some point between the two, a type piracy appeared in SparseArrays which caused the ambiguity.


test/derivatives/ArrayFunctionTests.jl
```diff
@@ -27,25 +27,42 @@ function testcat(f, args::Tuple, type, kwargs=NamedTuple())
...
         @assert length(args) == 2
-        x = f(track(args[1]), args[2]; kwargs...)
-        @test x isa type
-        @test value(x) == f(args...; kwargs...)
 
-        x = f(args[1], track(args[2]); kwargs...)
-        @test x isa type
-        @test value(x) == f(args...; kwargs...)
+        broken = f == hcat && (args[2] isa AbstractMatrix)
+        if broken && VERSION >= v"1.4"
```
devmotion, line 32: Why 1.4? AFAICT (and as mentioned eg in #259) this issue is caused by a type piracy in SparseArrays introduced in Julia 1.10 (see https://github.com/JuliaSparse/SparseArrays.jl/issues/431).

devmotion (commented):
While I appreciate "fixing" CI, I think this PR (and others merged recently) should have been reviewed.

gdalle:
> While I appreciate "fixing" CI, I think this PR (and others merged recently) should have been reviewed.

That's true, but the last issue I opened, outlining a genuine bug (#251) got zero answer in four months. I was unsure if anyone was actually watching or interested.

test/derivatives/ArrayFunctionTests.jl
gdalle: I didn't do the bisection to figure out which version of Julia was guilty. I just took the version that was above 1.3, on which I was sure the tests were passing.
Tbh we should just drop 1.0 and 1.3, even Julia doesn't support those anymore

gdalle:
Another reason why this PR was necessary is that for lack of a `@testset`, tests were failing and exiting _at the first failure_. So having a systematic failure in the concatenation tests on 1.10 means we never would have caught a bug happening after that one.

PR #273, opened 2025-04-05
Skip the exponent function

ChrisRackauckas:
It is a Base function that returns a scalar.

https://github.com/SciML/DiffEqBase.jl/blame/master/src/internal_itp.jl#L46

https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/14272335727/job/40008186966?pr=2662#step:6:1251

@oscardssmith

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReverseDiff"
 uuid = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
-version = "1.16.0"
+version = "1.16.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
The test seems to already be generated from the function list.

devmotion:
Looks good to me, it's also consistent with ForwardDiff: https://github.com/JuliaDiff/ForwardDiff.jl/blob/f00147ebd3615497b24bafb71c1f037857ca9408/src/dual.jl#L339 `cat` tests are failing (not broken anymore) on more recent Julia versions (someone should check what was changed when upstream to correctly mark them as not broken anymore...), otherwise tests pass on master.

devmotion (approved):

ChrisRackauckas:
Confirmed https://github.com/JuliaDiff/ReverseDiff.jl/actions/runs/14282119048/job/40032870650?pr=273#step:6:321

PR #290, opened 2026-09-14
Don't error when broadcasting over types without a definite field count

### LINKED ISSUE #289: "type does not have a definite number of fields"
```julia
using ReverseDiff
f(x) = sum(convert.(Float64, x) .* [2.0, 3.0])
ReverseDiff.gradient(f, [0.3, 0.5])
# ArgumentError: type does not have a definite number of fields
```

Claude said the following:

Root cause in ReverseDiff `src/derivatives/broadcast.jl`:

- Line 22: `istypeorclosure(::Base.RefValue{F}) where {F} = _istypeorclosure(F)` — when `convert.(T, x)` is broadcast, Julia wraps the scalar `T` (which is `Float64`, i.e. `Type{Float64}`) in a `RefValue{Type{Float64}}`, so `F = Type{Float64}`.
- Line 27: `@generated _istypeorclosure(::Type{F}) where {F} = :($(fieldcount(F) > 0))` — this calls `fieldcount(Type{Float64})`, which throws because Type is abstract and doesn't have a definite number of fields.

The fix should be in ReverseDiff: `_istypeorclosure` needs a guard for when `F` is a `Type`. Line 30 already has `mayhavetracked(b::Type) = false`, but the `RefValue` dispatch on line 22 bypasses that and goes straight to `_istypeorclosure`.

```julia
julia> versioninfo()
Julia Version 1.13.0
Commit d1c37793dd2 (2026-09-09 19:00 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: macOS (arm64-apple-darwin25.6.0)
  CPU: 15 × Apple M5 Pro
  WORD_SIZE: 64
  LLVM: libLLVM-20.1.8 (ORCJIT, apple-m1)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 15 virtual cores)
```

(It's not 1.13 specific though, the same error occurs on 1.12)

devmotion:
Fixes #289.

`_istypeorclosure` is `@generated`, so its `fieldcount(F)` runs at code-generation time and throws for types without a definite number of fields. Broadcasting `convert.(Float64, x)` wraps `Type{Float64}` in a `Base.RefValue` and hits exactly that, failing with `ArgumentError: type does not have a definite number of fields` before the broadcast is ever evaluated.

Answering `true` there takes the generic fallback. That's the conservative choice — `false` would unlock the fast paths that evaluate the broadcast function on `ForwardDiff.Dual` numbers, which isn't sound for values that may hide tracked numbers. `convert.(Float64, x)` now reports the informative error from `tracked.jl` suggesting `ReverseDiff.value`, and `convert.(Real, x)`, `x .* Any[...]` and `x .* Base.RefValue{Any}(...)` differentiate correctly. No path that worked before changes behaviour: every input whose answer changes used to throw.

The second commit is unrelated to #289 but needed for a green suite. The `cat` tests' `v"1.4" <= VERSION < v"1.11"` guard reports `Unexpected Pass` on Julia 1.6 - 1.9. The ambiguity they cover comes from the `hcat`/`vcat` methods SparseArrays pirates for `Union{AbstractVecOrMat{<:Number},Number}`, so it needs both SparseArrays >= 1.10, where that signature was widened, and SparseArrays to be loaded, which no longer happens transitively on 1.11+. Measured on 1.6 - 1.12: only 1.10 is affected, and there the failing assertions are exactly the ones already marked broken.

Third commit bumps the version to `v1.17.1`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


penelopeysm:
Thanks :)

